PR #6, opened 2016-05-11
Enable Travis-CI for Pilot

petricm:
Added the YML file for Travis, configured so that simultaneous 2.6 and 2.7 are tested. Sadly Travis only supports python >=2.6 otherwise we could also test older versions. 

Had to change PilotLogger since it used dict comprehension and adopt test to be compatible with 2.6

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,14 @@
+language: python
+python:
+  - "2.3"
+  - "2.4"
+  - "2.5"
+  - "2.6"
+  - "2.7"
+# command to install dependencies
+install: "pip install -r requirements.txt"
+# command to run tests
+script:
+  - export PYTHONPATH=${PWD%/*}
+  - ls $PYTHONPATH
+  - py.test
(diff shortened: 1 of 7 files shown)


petricm:
Forgot to mention. Go to https://travis-ci.org/profile/DIRACGrid and enable Travis for Pilot

fstagni:
Review OK

fstagni:
Great, already working and passing the tests: https://travis-ci.org/DIRACGrid/Pilot

Thanks a lot!

petricm:
Who is taking care, that the things are also merged into DIRAC/Pilot? Is Pilot the main development point and things are merged to DIRAC/Pilot or is the the other way around?

fstagni:
Pilot is for new developments. DIRAC/pilot is still "the production 
pilot" and so it will be for v6r15. Fixes to DIRAC/pilot should be 
ported also to Pilot (hopefully there won't be too many)

PR #7, opened 2016-05-12
Pilot loggers

wkrzemien:
This part covers some integration tests for Pilot Loggers and rabbitmq server. This part contains the basic pilot logger tests, with several utilitis.


wkrzemien:
99% of the problems are related to the files I included from DIRAC (not my code), and that are going to be probably modified anyway, so I would suggest to merge now unless you have some concrete remarks

PR #8, opened 2016-05-18
Enable coveralls and Landscape

petricm:
The same again for Pilot


petricm:
The same as before for the other repo.

PR #21, opened 2016-11-18
Add NagiosProbes to pilotCommands.py

Andrew-McNab-UK:
This PR adds the ability to run scripts following the Nagios Plugin API https://assets.nagios.com/downloads/nagioscore/docs/nagioscore/3/en/pluginapi.html within the Pilot. Scripts of this form are used as the worker node probes of the WLCG SAM and ETF monitoring of sites. The list of scripts can be set in Operations/SETUP/Pilot/NagiosProbes in the CS. The result of the probes is logged and can optionally be uploaded via HTTPS PUT to a URL given by NagiosPutURL in the CS. A corresponding consumer Nagios probe is in pre-production on the ETF services machines. For LHCb, the plan is use the Pilot Logging mechanism when this is ready to log the results of the probes and to provide ETF with a Nagios probe to query the pilot logging database somehow, via an intermediary service. This PR has been tested at 5 VM-based sites.


Pilot/user_data
```diff
@@ -193,35 +194,56 @@ cp /root/hostkey.pem /scratch/plt/etc/grid-security/hostkey.pem
...
+  done
 done
-) 2>&1 >/var/spool/joboutputs/useradd.log &
 
 cd /scratch/plt
 # Fetch the DIRAC pilot scripts
-if [ '##user_data_option_dirac_pilot_url##' == '' ] ; then
-  wget --no-directories --recursive --no-parent --execute robots=off --reject 'index.html*' https://lhcbproject.web.cern.ch/lhcbproject/Operations/diracpilot/
+if [ '##user_data_option_dirac_pilot_url##' != '' ] ; then
+  wget --no-directories --recursive --no-parent --execute robots=off --reject 'index.html*' --ca-directory=/etc/grid-security/certificates '##user_data_option_dirac_pilot_url##'
```
fstagni: I think you can reject '*.html'

Pilot/user_data
```diff
@@ -193,35 +194,56 @@ cp /root/hostkey.pem /scratch/plt/etc/grid-security/hostkey.pem
...
+elif [ '##user_data_url##' != '' ] ; then
+  # Remove user_data file name back to final slash
+  user_data_dir=`echo '##user_data_url##' | sed 's:[^/]*$::'`
+  wget --no-directories --recursive --no-parent --execute robots=off --reject 'index.html*' --ca-directory=/etc/grid-security/certificates "$user_data_dir"
```
fstagni: The ca directory should be `$X509_CERT_DIR`, if this variable is set. In practice it should be the same logic that we implemented in https://github.com/DIRACGrid/DIRAC/blob/integration/Core/scripts/dirac-install.py#L1333.

For LHCb, we should also consider other candidates, like in https://gitlab.cern.ch/lhcb-dirac/LHCbPilot/blob/master/LHCbPilot/LHCbPilotCommands.py#L136

Pilot/pilotCommands.py
```diff
@@ -1100,3 +1101,111 @@ def execute( self ):
...
+        pass
+
+    self.log.debug( 'NAGIOS PROBES [%s]' % ', '.join( self.nagiosProbes ) )
+
+  def _runNagiosProbes( self ):
+    """ Run the probes one by one
+    """
+
+    for probeCmd in self.nagiosProbes:
+     self.log.debug( "Running Nagios probe %s" % probeCmd )
```
fstagni: The indentation here is wrong. Please run pylint or follow what is in https://landscape.io/github/DIRACGrid/Pilot/pulls/21

fstagni:
There's no unit test. Can you please add one for the new command?

Pilot/user_data
Andrew-McNab-UK: I'll set $X509_CERT_DIR earlier in the user_data shell script but it's all hard coded in effect because the user_data itself is setting everything up. I agree within the pilot itself we should (and do) use $X509_CERT_DIR

schitic:
Test this please

schitic:
test this please

schitic:
test this please

schitic:
test this please

schitic:
add to whitelist

Andrew-McNab-UK:
I've updated user_data, pilotCommands.py and added a test. Then tested this on a VM at a site and verified I've not broken anything in the process. Could you accept the PR now. Thanks, Andrew

fstagni:
OK now. Merging it.

PR #36, opened 2016-12-09
Handle Commands/Extensions as proper lists or comma lists

Andrew-McNab-UK:
This is to get ready for a corresponding change in the pilotSynchronizer to publish all multivalue options in pilot.json as comma-list strings rather some as proper lists, inconsistently. If other projects add pilot commands and corresponding multivalue options in the CS, then everything will then be consistent. It also avoids the need to hardcode multivalue->list processing in the pilotSynchronizer for some options (Commands and CommandExtensions).

There are also a couple of changes to the temporary LHCb code in the user_data file which uploads log files to lhcb-depo.cern.ch until we can use the extended pilot logging for production.

### DIFF
--- a/Pilot/user_data
+++ b/Pilot/user_data
@@ -277,11 +277,12 @@ cp -f /scratch/plt/jobagent.*.log /scratch/plt/shutdown_message* /var/log/boot.l
   for i in *
   do
    if [ -f $i ] ; then 
-    curl --capath $X509_CERT_DIR --cert /root/hostkey.pem --cacert /root/hostkey.pem --location --upload-file "$i" "$JOBOUTPUTS/"
+    curl --capath $X509_CERT_DIR --cert /root/hostkey.pem --cacert /root/hostkey.pem --location --upload-file "$i" \
+     "$JOBOUTPUTS/"
 
     # This will be replaced by extended pilot logging??
     curl --capath $X509_CERT_DIR --cert /root/hostkey.pem --cacert /root/hostkey.pem --location --upload-file "$i" \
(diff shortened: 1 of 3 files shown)


Pilot/user_data
```diff
@@ -277,11 +277,12 @@ cp -f /scratch/plt/jobagent.*.log /scratch/plt/shutdown_message* /var/log/boot.l
...
   do
    if [ -f $i ] ; then 
-    curl --capath $X509_CERT_DIR --cert /root/hostkey.pem --cacert /root/hostkey.pem --location --upload-file "$i" "$JOBOUTPUTS/"
+    curl --capath $X509_CERT_DIR --cert /root/hostkey.pem --cacert /root/hostkey.pem --location --upload-file "$i" \
+     "$JOBOUTPUTS/"
 
     # This will be replaced by extended pilot logging??
     curl --capath $X509_CERT_DIR --cert /root/hostkey.pem --cacert /root/hostkey.pem --location --upload-file "$i" \
-      "https://lbvobox200.cern.ch:9132/hosts/$CE_NAME/$HOSTNAME/$VM_UUID/"
+     "https://lhcb-depo.cern.ch:9132/hosts/##user_data_space##/##user_data_machinetype##/##user_data_machine_hostname##/##user_data_uuid##/"
```
fstagni: This file is still LHCb specific and it's here only for convenience, but sooner or later should be made generic. 

What's lhcb-depo?

Andrew-McNab-UK:
Thanks. As soon as we can use the extended pilot logging in production, then yes the requirement to put the logs on the HTTPS fileserver lhcb-depo.cern.ch (=lbvobox200 at the moment) goes away. We can get all the pilot log files on all platforms in one place then.

PR #38, opened 2016-12-17
Add test execution time to saved Nagios probe results

Andrew-McNab-UK:
The one-line change to pilotCommands.py puts the time the Nagios probe was run in the saved results file itself. 

This is needed by check_pilot_results in LHCbSAM, which needs to know the time so it can flag stale results as "unknown".  This mechanism allows the test result files to be migrated or mirrored from one HTTPS server to another without needing to update the file datestamps reported by HTTP Last-Modified header.

### DIFF
--- a/Pilot/pilotCommands.py
+++ b/Pilot/pilotCommands.py
@@ -1192,7 +1192,7 @@ class NagiosProbes( CommandBase ):
                                                 key_file  = os.environ['X509_USER_PROXY'],
                                                 cert_file = os.environ['X509_USER_PROXY'] )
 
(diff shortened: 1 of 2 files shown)


fstagni:
IIUC this format is understood here: https://gitlab.cern.ch/lhcb-dirac/LHCbSAM/commit/8fb87f2299c0704f862196f19c9ab59f403994c7

I just want confirmation that this is safe to merge (otherwise I'll wait for new year)

Andrew-McNab-UK:
Yes, it's ready to merge and tested on several sites.

PR #39, opened 2017-01-23
--commandOptions and UnpackDev command 

Andrew-McNab-UK:
As discussed a few minutes ago, this code adds the option --commandOptions and the Pilot command UnpackDev.

--commandOptions/-Z creates a dictionary from the value of this option which is intended to be used by locally written pilot commands which need command line options. This is to avoid needing to fork pilotTools.py when using such a pilot command. The option given to --commandOptions is split by comma and then colon to create the dictionary. For example --commandOptions=a:1,b:2,c:3 results in the dictionary self.commandOptions = { 'a' : '1', 'b' : '2', 'c' : '3' } being available to the pilot command classes.

The pilot command UnpackDev is added to pilotCommands.py and unpacks the file dev.tgz in the pilot directory if the command is executed. This enables testing of development versions of DIRAC etc within pilots run on the grid. Typically one or two Python files will be included in dev.tgz so they are found on the Python path before the release version.


Pilot/pilotCommands.py
```diff
@@ -1210,3 +1210,33 @@ def execute( self ):
...
+
+  def __init__( self, pilotParams ):
+    """ c'tor
+    """
+    super( UnpackDev, self ).__init__( pilotParams )
+    self.devFile = 'dev.tgz'
+
+  def execute( self ):
+    """ Standard entry point to a pilot command
+    """
```
fstagni: I suggest that we execute what's here iff `os.path.exists(self.devFile)`

Pilot/pilotCommands.py
```diff
@@ -1210,3 +1210,33 @@ def execute( self ):
...
+    self.log.info( 'Adding %s to bashrc' % os.getcwd() )
+    retCode, output = self.executeAndGetOutput( 'sed -i "s,^PYTHONPATH=,PYTHONPATH=%s:," bashrc' % os.getcwd() )
+    self.log.info( output, header = False )
+            
+    self.log.info( 'Unpacking ' + self.devFile )
+    retCode, output = self.executeAndGetOutput( 'tar zxvf ' + self.devFile )
```
fstagni: I think using `tarfile` https://docs.python.org/2/library/tarfile.html is much cleaner!

Pilot/pilotTools.py
```diff
@@ -488,8 +489,9 @@ def __init__( self ):
     self.cmdOpts = ( ( 'b', 'build', 'Force local compilation' ),
                      ( 'd', 'debug', 'Set debug flag' ),
                      ( 'e:', 'extraPackages=', 'Extra packages to install (comma separated)' ),
-                     ( 'E:', 'commandExtensions=', 'Python module with extra commands' ),
-                     ( 'X:', 'commands=', 'Pilot commands to execute commands' ),
+                     ( 'E:', 'commandExtensions=', 'Python modules with extra commands' ),
+                     ( 'X:', 'commands=', 'Pilot commands to execute' ),
+                     ( 'Z:', 'commandOptions=', 'Options parsed by command modules' ),
```
fstagni: TBH I am a bit tempted to remove the "1 letter switch" and leave only the long one.

Pilot/tests/Test_Pilot.py
```diff
@@ -27,10 +27,10 @@ def setUp( self ):
...
 
-    sys.argv[1:] = ['--Name', 'grid1.example.com']
+    sys.argv[1:] = ['--Name', 'grid1.example.com', '--commandOptions', 'a=1,b=2', '-Z', 'c=3' ]
 
     self.pp = PilotParams()
-
+    
   def tearDown( self ):
     try:
       os.remove( 'pilot.json' )
```
fstagni: add 'dev.tgz' to the list of files to remove.

Pilot/tests/Test_Pilot.py
```diff
@@ -27,13 +27,15 @@ def setUp( self ):
...
+      os.remove( 'dev.tgz' )
+      os.remove( '123.txt' )
     except IOError:
```
fstagni: It's OSError to be catched, not IOError

Pilot/pilotCommands.py
Andrew-McNab-UK: I'd quite like an error if the command is run but the file isn't supplied. I don't want to be thinking the changes have been applied when they've not.

fstagni:
Up!

PR #40, opened 2017-02-10
added Generic option

fstagni:


fstagni:
This is included in #41

PR #41, opened 2017-03-10
Using bashrc for environment (taken from production version - DIRAC v6r17)

fstagni:


fstagni:
@Andrew-McNab-UK and @CinziaLu please have a look (little time, sorry!)

PR #46, opened 2017-08-11
Tags in Pilot3

Andrew-McNab-UK:
This patch adds 

- an option to locally define tags when customising the user_data file, which is useful for testing and for features the site knows about at the host level (e.g. GPUs). 

- RequiredTag= is added back in alongside Tag=. This was lost when this functionality was moved from the SiteDirector to the pilotCommands.py in DIRAC (added in PR #2823 and (re)moved in PR #3348)

This functionality has been verified to work in VMs: payloads with Tags=xxxx match and run; payloads without Tags=xxxx are not run.

### DIFF
--- a/Pilot/pilotCommands.py
+++ b/Pilot/pilotCommands.py
@@ -385,7 +385,8 @@ class CheckCECapabilities( CommandBase ):
       if self.debugFlag:
         self.cfg.append( '-ddd' )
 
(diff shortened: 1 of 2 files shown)


fstagni:
It's not so clear to me if this PR is needed. Especially after https://github.com/DIRACGrid/DIRAC/pull/3441/files, which I just ported in #47. Would you please clarify it?

Andrew-McNab-UK:
Yes it's not needed now. Thanks for forward porting Andrei's changes.

PR #49, opened 2017-10-23
Changes to user_data_** and replace UnpackDev with ReplaceDIRACCode

Andrew-McNab-UK:
This PR has updates to user_data_vm and adds user_data_sc (for Singularity containers and ok with Docker too).

In pilotCommands.py, the UnpackDev command is removed and a modified version of ReplaceDIRACCode from Pilot 2.0 is used instead. The main change is to use TGZ instead of ZIP, as the ZipFile module does not preserve the execute bit on scripts.

I've tidied up pilotTools.py, in particular ordering the list of options alphabetically, which identified one duplicate ("-C") marked with FIXME. I also removed a duplicate of the __initCommandLine2() function.

The functionality of Pilot 3.0 is otherwise unchanged.

### DIFF
--- a/Pilot/pilotCommands.py
+++ b/Pilot/pilotCommands.py
@@ -288,8 +288,10 @@ class ReplaceDIRACCode( CommandBase ):
 
     # Add the ReplacementCode directory to the Python path
     self.pp.installEnv['PYTHONPATH'] = os.getcwd() + os.path.sep + 'ReplacementCode' + ':' + self.pp.installEnv['PYTHONPATH']
+    self.pp.installEnv['PATH'] = os.getcwd() + os.path.sep + 'ReplacementCode' + os.path.sep + 'scripts:' + self.pp.installEnv['PATH']
 
-    self.log.info( "TGZ file %s unpacked. PYTHONPATH updated to be %s" % ( self.pp.replaceDIRACCode, self.pp.installEnv['PYTHONPATH'] ) )
+    self.log.info( "TGZ file %s unpacked. PYTHONPATH updated to be %s and PATH updated to be %s" % 
+                         ( self.pp.replaceDIRACCode, self.pp.installEnv['PYTHONPATH'], self.pp.installEnv['PATH'] ) )
 
 class ConfigureBasics( CommandBase ):
   """ This command completes DIRAC installation.
--- a/Pilot/pilotCommands.py
+++ b/Pilot/pilotCommands.py
@@ -22,6 +22,7 @@ import os
 import time
 import stat
 import socket
+import urllib
 import tarfile
 import httplib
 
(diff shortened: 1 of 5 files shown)


Pilot/user_data_sc
```diff
@@ -0,0 +1,116 @@
+#!/bin/sh
```
fstagni: add some doc.

Pilot/tests/Test_Pilot.py
```diff
@@ -109,18 +109,6 @@ def test_NagiosProbes ( self ):
     self.assertEqual( nagios.nagiosProbes, ['Nagios1', 'Nagios2'] )
     self.assertEqual( nagios.nagiosPutURL, 'https://127.0.0.2/' )
 
-  def test_UnpackDev ( self ):
```
fstagni: test_replaceDirac?

Pilot/pilotTools.py
```diff
@@ -488,43 +489,44 @@ def __init__( self ):
...
-                     ( 'u:', 'url=', 'Use <url> to download tarballs' ),
                      ( 'r:', 'release=', 'DIRAC release to install' ),
-                     ( 'n:', 'name=', 'Set <Site> as Site Name' ),
+                     ( 's:', 'section=', 'Set base section for relative parsed options' ),
+                     ( 'u:', 'url=', 'Use <url> to download tarballs' ),
+                     ( 'x:', 'execute=', 'Execute instead of JobAgent' ),
+                     ( 'y:', 'CEType=', 'CE Type (normally InProcess)' ),
+                     ( 'z',  'pilotLogging', 'Activate pilot logging system' ),
+                     ( 'C:', 'configurationServer=', 'Configuration servers to use' ), # FIXME: -C is used twice! Which one to keep?
+                     ( 'C:', 'certLocation=', 'Specify server certificate location' ), # FIXME: -C is used twice! Which one to keep?
```
fstagni: use `-K`

Pilot/pilotTools.py
```diff
@@ -488,43 +489,44 @@ def __init__( self ):
...
+                     ( 'o:', 'option=', 'Option=value to add' ),
                      ( 'p:', 'platform=', 'Use <platform> instead of local one' ),
-                     ( 'u:', 'url=', 'Use <url> to download tarballs' ),
                      ( 'r:', 'release=', 'DIRAC release to install' ),
-                     ( 'n:', 'name=', 'Set <Site> as Site Name' ),
+                     ( 's:', 'section=', 'Set base section for relative parsed options' ),
+                     ( 'u:', 'url=', 'Use <url> to download tarballs' ),
+                     ( 'x:', 'execute=', 'Execute instead of JobAgent' ),
+                     ( 'y:', 'CEType=', 'CE Type (normally InProcess)' ),
+                     ( 'z',  'pilotLogging', 'Activate pilot logging system' ),
```
fstagni: do you by chance know why some of them has an `=` and some not? e.g.:
```
pilotLogging
```
and
```
CEType=
```

does it make a difference? 

Similarly, for the "one letter options: some have `:` and some not... does it make a difference?

fstagni: Of course! Thanks.

Pilot/pilotCommands.py
```diff
@@ -654,6 +697,14 @@ def __setFlavour( self ):
       if 'BOINC_HOST_NAME' in os.environ:
         self.boincHostName = os.environ['BOINC_HOST_NAME']
 
+    # Pilot reference is given explicitly in environment
+    if 'PILOT_UUID' in os.environ:
+      pilotRef = os.environ['PILOT_UUID']
```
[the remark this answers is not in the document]
Andrew-McNab-UK: That will set pilotRef to None if the environment variable isn't present. Ok for PILOT_UUID but not for some of the batch/grid systems which the code is trying to identify one by one in an order that matters (IIUC). I think this code should be moved to pilotTools where both pilot commands and pilot logging can use it. It can be tidied up at that point, but it still has to determine the flavour there too, using the logic currently in this function.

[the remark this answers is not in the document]
fstagni: Right, so better `pilotRef = os.environ.get('PILOT_UUID', pilotRef)`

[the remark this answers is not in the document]
fstagni: Can you please either do the move yourself, or add the above comment directly in the code?

fstagni:
I will merge this tomorrow. Please just review the other trivial PR I just made.

PR #52, opened 2018-01-19
(for tests) sourcing DIRAC/tests/Jenkins/utilities.sh

fstagni:


fstagni:
This is really easy... so I merge directly.

PR #59, opened 2018-07-30
Fixes for multi-core support -- backports #3739

fstagni:
Portings from https://github.com/DIRACGrid/DIRAC/pull/3739


fstagni:
Merging very soon. A review would be better.

fstagni:
Ported in https://github.com/DIRACGrid/Pilot/pull/59

PR #60, opened 2018-11-12
just debug

fstagni:


Pilot/dirac-pilot.py
```diff
@@ -18,48 +18,47 @@
...
 
-__RCSID__ = "$Id$"
-
 if __name__ == "__main__":
 
-  pilotStartTime = int( time.time() )
+  pilotStartTime = int(time.time())
 
-  log = Logger( 'Pilot' )
+  log = Logger('Pilot', debugFlag=True)
```
wkrzemien: This is maybe question of convention, but for me debug should not be true by default.

fstagni:
Fixed, please re-review.

PR #66, opened 2019-03-03
Add PilotLogger version that was somehow lost in merging

wkrzemien:
Refactored PilotLogger  version that was somehow overwritten by older versions. 
Also, add config file with autopep8 settings.

### DIFF
--- a/Pilot/tests/Test_PilotLogger.py
+++ b/Pilot/tests/Test_PilotLogger.py
@@ -1,60 +1,109 @@
 """ Unit tests for PilotLogger
 """
 
-# pylint: skip-file
+# pylint: disable=protected-access, missing-docstring, invalid-name, line-too-long
 
+import unittest
 import os
-
-import pytest
-
-from Pilot.PilotLogger import PilotLogger, getPilotUUIDFromFile
+from Pilot.PilotLogger import PilotLogger, getPilotUUIDFromFile, addMissingConfiguration
 from Pilot.PilotLoggerTools import getUniqueIDAndSaveToFile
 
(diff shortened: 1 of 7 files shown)


Pilot/tests/Test_PilotLogger.py
```diff
@@ -1,60 +1,115 @@
...
-def test_failureEmpty(rmFiles):
-  logger = PilotLogger()
-  assert logger._isCorrectStatus('') is False
+if __name__ == '__main__':
+  suite = unittest.defaultTestLoader.loadTestsFromTestCase(TestGetPilotUUIDFromFile)
+  suite.addTest(unittest.defaultTestLoader.loadTestsFromTestCase(TestPilotLogger_isCorrectStatus))
+  suite.addTest(unittest.defaultTestLoader.loadTestsFromTestCase(TestPilotLogger_init))
+  suite.addTest(unittest.defaultTestLoader.loadTestsFromTestCase(TestPilotLogger_addMissingConfiguration))
+  suite.addTest(unittest.defaultTestLoader.loadTestsFromTestCase(TestPilotLogger_sendMessage))
+  testResult = unittest.TextTestRunner(verbosity=2).run(suite)
```
fstagni: Please, as last lines of these integration tests (all those in Pilot/tests) that are coded using python `unittest`, use:

```
  sys.exit(not testResult.wasSuccessful())
```

This way it will exit with return code !=0 if not successful (it's useful in Jenkins and other CI tools).

This is true for this file and the others in this dir too.

wkrzemien:
@fstagni This one is passing all the working tests. So maybe you could merge it if you don't have more comments?

PR #67, opened 2019-03-27
dirac-install: do not set REQUESTS_CA_BUNDLE in bashrc

chaen:
port changes from https://github.com/DIRACGrid/DIRAC/pull/3992
@fstagni I've done it against devel, not sure that this is the correct branch though


fstagni:
This PR is for `devel` branch, while it should be for `master`. Sorry I didn't notice.

PR #69, opened 2019-03-31
Remove Python 2.6 tests from .travis.yml

wkrzemien:
It was not working anyway, so better not to add the noise.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,5 @@
 language: python
 python:
-  - "2.6"
   - "2.7"
 
 env:


chaen:
Hum, not sure about that. 
The pilot MUST run on 2.6, so the code/tests should rather be fixed

wkrzemien:
I agree. This ment to be rather provocative to try to fix it ASAP, cause currently even master is not passing and it is like that since February.

fstagni:
If we are going to remove this check, than it should be removed starting from the `master` branch. Anyway, the reason why this fails is because some package stopped supporting python 2.6. If you look at the log https://travis-ci.org/DIRACGrid/Pilot/jobs/506286740 it stops at PyYAML, so it's a matter of adapting the requirements.txt file for this case.

PR #71, opened 2019-04-09
remove python-coveralls

fstagni:


fstagni:
This is what was failing the tests for python 2.6. Now everything should be green. Merging propagating. 
@wkrzemien

PR #77, opened 2019-06-13
Added option to install with DIRACOS

fstagni:
Including the tests


fstagni:
Review, please, so I merge.

PR #79, opened 2019-06-25
Payload processors

Andrew-McNab-UK:
This PR adds support for multiprocessor payloads in DIRAC VMs defined by user_data_vm. In the pilotTools/Commands, self.pp.payloadProcessors and self.pp.pilotProcessors are introduced which are available to other pilot commands (eg MultiLaunchAgent). Required tags and normal tags are now stored in CEDefaults by CheckWN rather than by CheckCE to allow MultiProcessor and NProcessors tags to be created when they have not been created as part of the pilot definition (eg inside a VM.) The MultiLaunchAgent now launches int(P/p) job agents where P is the number of processors allocated to the pilot (or VM) and p is the number of processors per payload.  

In CheckWN, self.cfg is used throughout instead of a mixture of cfg and self.cfg (which I think was just there for historical reasons?)

### DIFF
--- a/Pilot/pilotCommands.py
+++ b/Pilot/pilotCommands.py
@@ -397,18 +397,9 @@ class CheckCECapabilities(CommandBase):
     if resourceDict.get('Tag'):
       self.pp.tags += resourceDict['Tag']
 
-    self.pp.tags = list(set(self.pp.tags))
-    if self.pp.tags:
-      cfg.append('-o "/Resources/Computing/CEDefaults/Tag=%s"' % ','.join((str(x) for x in self.pp.tags)))
-
     # RequiredTags are like Tags.
     if resourceDict.get('RequiredTag'):
       self.pp.reqtags += resourceDict['RequiredTag']
-
-    self.pp.reqtags = list(set(self.pp.reqtags))
-    if self.pp.reqtags:
-      cfg.append('-o "/Resources/Computing/CEDefaults/RequiredTag=%s"' %
-                 ','.join((str(x) for x in self.pp.reqtags)))
 
     # LocalCE type for singularity
     if resourceDict.get('Container') in ["Singularity", "singularity"]:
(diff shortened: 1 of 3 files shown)


fstagni:
Hi Andrew, this PR is adding more support for MultiLaunchAgent, but I don't like much this command.
I think that the correct solution would instead be in using another CE (PoolCE) instead of "InProcessCE". The PoolCE solution is coded and should just be tried out.

Andrew-McNab-UK:
I don't have time to do that unfortunately, and this PR is working now. Given that MultiLaunchAgent is there already (and relied on in production), I don't think there's any harm in adding these small changes to it. If someone has time to validate the PoolCE for VMs with sudo or singularity isolation etc etc, then I'd be happy to see the MultiLaunchAgent removed.

fstagni:
Integration test ran fine (https://jenkins-dirac.web.cern.ch/blue/organizations/jenkins/Pilot3_pipeline_testJob/detail/Pilot3_pipeline_testJob/9/pipeline).

I have added https://github.com/DIRACGrid/Pilot/issues/86 to remember to remove the MultiLaunchAgent solution in the future.

fstagni:
@wkrzemien please also have a look at this, and if you don't see any issue we merge it.

After, the `devel` branch will be merged in the `master` branch.

Pilot/pilotCommands.py
```diff
@@ -487,19 +482,29 @@ def execute(self):
...
     # values are preferred
     if "WholeNode" in self.pp.tags:
-      numberOfProcessors = numberOfProcessorsOnWN
+      self.pp.payloadProcessors = numberOfProcessorsOnWN
     if self.pp.maxNumberOfProcessors > 0:
-      numberOfProcessors = min(numberOfProcessorsOnWN, self.pp.maxNumberOfProcessors)
+      self.pp.payloadProcessors = min(numberOfProcessorsOnWN, self.pp.maxNumberOfProcessors)
 
-    if not numberOfProcessors:
+    if not self.pp.payloadProcessors:
```
wkrzemien: I think this can never happen, cause you explicitely set this variable in 475

wkrzemien: but it does no harm

Andrew-McNab-UK: I think it can happen if dirac-wms-get-wn-parameters returns zero processors. (It is the same as the current code anyway.)

PR #89, opened 2019-07-16
Enhance REST pilot logger 

wkrzemien:
Small refactorization making  errors and warnings more verbose
Also, extend the CI tests


fstagni:
Please rebase.

JenkinsfilePilotLogger
```diff
@@ -0,0 +1,134 @@
+#!/usr/bin/env groovy
```
fstagni: This file is just for your convenience, so it shouldn't be added to this PR.

PR #90, opened 2019-07-18
Remove option to run pilot loggers from tests

wkrzemien:
Apparently pilotLogger flag was on in the tests. By default it was trying to send messages to local file.
I turned that off.
Also,  the stomp/requests function is not necessary in that case


wkrzemien:
ok.

PR #91, opened 2019-07-18
Setting pilotProcessors to what returned by dirac-wms-get-wn-parameters

fstagni:
Used for MP jobs. This goes together with https://github.com/DIRACGrid/DIRAC/pull/4175

### DIFF
--- a/Pilot/pilotCommands.py
+++ b/Pilot/pilotCommands.py
@@ -387,13 +387,6 @@ class CheckCECapabilities(CommandBase):
     if "WholeNode" in resourceDict:
       self.pp.tags.append('WholeNode')
 
-    # If MaxNumberOfProcessors not defined check for NumberOfProcessors
-    if self.pp.maxNumberOfProcessors == 0:
-      self.pp.maxNumberOfProcessors = int(
-          resourceDict.get(
-              "MaxNumberOfProcessors", resourceDict.get(
-                  "NumberOfProcessors", 0)))
-
      # Tags must be added to already defined tags if any
     if resourceDict.get('Tag'):
       self.pp.tags += resourceDict['Tag']
@@ -491,8 +484,6 @@ class CheckWNCapabilities(CommandBase):
     self.pp.payloadProcessors = 1
     if "WholeNode" in self.pp.tags:
       self.pp.payloadProcessors = self.pp.pilotProcessors
-    if self.pp.maxNumberOfProcessors > 0:
-      self.pp.payloadProcessors = min(self.pp.pilotProcessors, self.pp.maxNumberOfProcessors)
 
     self.log.info('pilotProcessors = %d' % self.pp.pilotProcessors)
     self.log.info('payloadProcessors = %d' % self.pp.payloadProcessors)
--- a/Pilot/pilotCommands.py
+++ b/Pilot/pilotCommands.py
@@ -478,17 +478,20 @@ class CheckWNCapabilities(CommandBase):
     # pilotProcessors is basically the number of processors this pilot is "managing"
     self.pp.pilotProcessors = numberOfProcessorsOnWN
 
-    # payloadProcessors is the number of processors used by the single payloads.
+    # payloadProcessors is the max number of processors used by the single payloads.
     # We store payloadProcessors in the global parameters so that other commands can more easily use it.
-    # (MultiLaunchAgent is right now is the only consumer)
+    # (MultiLaunchAgent is right now the only consumer)
     self.pp.payloadProcessors = 1
     if "WholeNode" in self.pp.tags:
       self.pp.payloadProcessors = self.pp.pilotProcessors
+    if self.pp.maxNumberOfProcessors > 0:
+      self.pp.payloadProcessors = min(self.pp.pilotProcessors, self.pp.maxNumberOfProcessors)
 
     self.log.info('pilotProcessors = %d' % self.pp.pilotProcessors)
     self.log.info('payloadProcessors = %d' % self.pp.payloadProcessors)
     self.cfg.append(
         '-o "/Resources/Computing/CEDefaults/NumberOfProcessors=%d"' % self.pp.pilotProcessors)
+        '-o "/Resources/Computing/CEDefaults/NumberOfPayloadProcessors=%d"' % self.pp.payloadProcessors)
 
     maxRAM = self.pp.queueParameters.get('MaxRAM', maxRAM)
     if maxRAM:
--- a/Pilot/pilotCommands.py
+++ b/Pilot/pilotCommands.py
@@ -471,10 +471,6 @@ class CheckWNCapabilities(CommandBase):
       self.log.error("Could not get resource parameters [ERROR %d]" % retCode)
       self.exitWithError(retCode)
 
-    # We store payloadProcessors in the global parameters so that other
-    # commands can more easily use it (eg MultiLaunchAgent, which right now is the only consumer)
-    self.pp.payloadProcessors = 1
-
     try:
       result = result.split(' ')
       numberOfProcessorsOnWN = int(result[0])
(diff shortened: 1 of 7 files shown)


fstagni:
@Andrew-McNab-UK please have a look at the changes I have done in this PR WRT multi-processor pilots.

Pilot/pilotCommands.py
```diff
@@ -484,27 +474,21 @@ def execute(self):
 
     # If NumberOfProcessors or MaxRAM are defined in the resource configuration, these
     # values are preferred
+
+    # pilotProcessors is basically the number of processors this pilot is "managing"
+    self.pp.pilotProcessors = numberOfProcessorsOnWN
```
Andrew-McNab-UK: Incidentally, this naming is wrong for historical reasons: the numberOfProcessorsOnWN is actually receiving the number of processors allocated to the job slot.

Pilot/pilotCommands.py
```diff
@@ -484,27 +474,21 @@ def execute(self):
...
-        self.pp.reqtags.append('MultiProcessor')
-      if ('%dProcessors' % self.pp.payloadProcessors) not in self.pp.tags:
-        self.pp.tags.append('%dProcessors' % self.pp.payloadProcessors)
+      self.pp.payloadProcessors = self.pp.pilotProcessors
 
+    self.log.info('pilotProcessors = %d' % self.pp.pilotProcessors)
     self.log.info('payloadProcessors = %d' % self.pp.payloadProcessors)
     self.cfg.append(
-        '-o "/Resources/Computing/CEDefaults/NumberOfProcessors=%d"' % self.pp.payloadProcessors)
+        '-o "/Resources/Computing/CEDefaults/NumberOfProcessors=%d"' % self.pp.pilotProcessors)
```
Andrew-McNab-UK: That's used in matching payloads (hence payloadProcessors). It mustn't just be the total number of processors the pilot has in case the pilot is going to run multiple payloads in parallel.

Pilot/pilotTools.py
```diff
@@ -618,8 +614,6 @@ def __initCommandLine2(self):
         self.installation = v
       elif o == '-p' or o == '--platform':
         self.platform = v
-      elif o == '-m' or o == '--maxNumberOfProcessors':
```
Andrew-McNab-UK: Removing --maxNumberOfProcessors will break multiprocessor support in grid and VM pilots, as they both use it on the dirac-pilot command line. In the VMs its used to distinguish between running 8 processor payloads or 8 single processor payloads.

Pilot/pilotCommands.py
```diff
@@ -484,27 +474,21 @@ def execute(self):
...
+    # pilotProcessors is basically the number of processors this pilot is "managing"
+    self.pp.pilotProcessors = numberOfProcessorsOnWN
+
+    # payloadProcessors is the number of processors used by the single payloads.
+    # We store payloadProcessors in the global parameters so that other commands can more easily use it.
+    # (MultiLaunchAgent is right now is the only consumer)
+    self.pp.payloadProcessors = 1
     if "WholeNode" in self.pp.tags:
-      self.pp.payloadProcessors = numberOfProcessorsOnWN
-    if self.pp.maxNumberOfProcessors > 0:
```
Andrew-McNab-UK: This allows payloadProcessors to be > 1 and is used to make the multiprocessor payloads work.

Pilot/pilotCommands.py
```diff
@@ -484,27 +474,21 @@ def execute(self):
...
-      self.log.warn("Could not retrieve number of processors, assuming 1")
-      self.pp.payloadProcessors = 1
-
-    # Make sure the multiprocessor tags are present if not already there
-    if self.pp.payloadProcessors > 1:
-      if 'MultiProcessor' not in self.pp.tags:
-        self.pp.tags.append('MultiProcessor')
-      if 'MultiProcessor' not in self.pp.reqtags:
-        self.pp.reqtags.append('MultiProcessor')
-      if ('%dProcessors' % self.pp.payloadProcessors) not in self.pp.tags:
```
Andrew-McNab-UK: After my PR the other week I realised that GridPP has 8Processors (or whatever) as the required tag and MultiProcessor as a normal tag. It would be ok to remove the tags part here and apply them from the dirac-pilot command line in the submitted job script or in the VM user_data.

fstagni:
So, we need to decide once and for all how to use these MP nodes. I am not talking about the MP queues, but about pilots started in a WN that needs to decide autonomously how to handle the processors.

So, let's start with some definitions. IMO:
- `pilotProcessors`: the number of processors this pilot is "managing". The slot. This has nothing to do with the size of the payloads (except that of course the size of the payloads can't be bigger)
- `payloadProcessors`: this is used only by MultiLaunchAgent (and should disappear with it) and I am not sure how to define it. Please give me a definition.
`maxNumberOfProcessors`: I also don't know how to define it, so please do. The SiteDirector never uses it, so I removed it.

IMO the only needed thing is the number of processors assigned to the slot. Then the matching can be done using the "Pool" of processors available.

Andrew-McNab-UK:
Following the PR earlier this month, the MultiLaunchAgent can run one or more multiprocessor payloads in parallel, rather than just one or more single processor payload. payloadProcessors is the number of processors each of these multiprocessor payloads has. That part is new. However, payloadProcessors itself is not new: it was there before as numberOfProcessors. maxNumberOfProcessors allows the number of processors allocated to the pilot to be set from the dirac-pilot command line. GridPP DIRAC uses this in production for multiprocessor grid pilot jobs eg by SKA and LSST. 

What I really want to avoid is a situation where the master Pilot code diverges from what I can deploy for the GridPP DIRAC VMs. 

So I would much prefer that you add/rename things to get the PoolCE working, then we check it works in production, including with pilot vs payload isolation somehow (Singularity?) and then we remove the MultiLaunchAgent and features only it uses. This means there would be an overlap period where both PoolCE and MultiLaunchAgent can be used. That would give time for GridPP DIRAC to change too.

As I said, I am happy to get rid of the MultiLaunchAgent, but I want to see an evolutionary approach which always provides production-validated functionality, and that requires an overlap.

fstagni:
So, IIUC `maxNumberOfProcessors` is the same as `pilotProcessors`. Can you confirm?

In general I agree with the strategy.

Andrew-McNab-UK:
Sorry, I should have said that "maxNumberOfProcessors allows the number of processors allocated to the pilot which the pilot can allocate to one payload to be set from the dirac-pilot command line. It is used to set payloadProcessors unless other limits are reached (like the number of processors on the WN)." I've added an Edited To Add note in case anyone reads that out of context.

fstagni:
Last week we merged in v6r22 this PR: https://github.com/DIRACGrid/DIRAC/pull/4175
Please have a look at it, the function `getNumberOfProcessors` which is called by `dirac-wms-get-get-parameters.py`, which is what's used here by the pilot (in `CheckWNCapabilities`) to know the number of processors available. This number is then inserted in `/Resources/Computing/CEDefaults/NumberOfProcessors`. This value HAS TO BE `pilotProcessors` (so the total number of processors on the WN), as this is what's always been meant for. 
As of today, instead, `/Resources/Computing/CEDefaults/NumberOfProcessors` is set to the value of `payloadProcessors`, which breaks the inProcess CE (in fact it works if and only if it's used inside the MultiLaunchAgent scenario).

My suggestion is to store `payloadProcessors` inside a new value, e.g. `/Resources/Computing/CEDefaults/NumberOfPayloadProcessors`. What do you think?

Andrew-McNab-UK:
Adding that new value looks clearer. 

What we're doing here is factorising things that were previously equivalent when we just had single processor payloads. How does having NumberOfProcessors = the payload size break the InProcess CE btw? In the LaunchAgent() with InProcessCE you've only got a single payload, so doesn't pilotProcessors=payloadProcessors? 

Can we also avoid saying "number of processors on the WN" when we mean "allocated to the pilot" as that will creep into the code again in situations other than WholeNode.

fstagni:
Sure we should always say `pilotProcessors` instead of "processors of the WN", I agree.

The thing is that the `/Resources/Computing/CEDefaults/NumberOfProcessors` has to be the `pilotProcessors`. This is what's used by PoolCE but also what's advocated in the documentation, so it can't be the `payloadProcessors`, because if we run a single JobAgent it's a different meaning.

I will update the code in this PR for a new evaluation.

fstagni:
Can you have a look now?

fstagni:
Up!

fstagni:
@Andrew-McNab-UK  ?

Andrew-McNab-UK:
Looks ok as far as I can see. Not had time to test it but I will raise a PR if anything comes up.

fstagni:
Merging in 30 mins or less.

PR #105, opened 2020-02-20
[WIP] Updated pilot logging

wkrzemien:
Pilot code which is a part of pilot logging with a two-stage scheme architecture:
Pilot ----> REST API (Tornado server) ---> MQServer

Tornado server:
1. checks validity of proxy
2. checks if the pilot's DN is on the list


.travis.yml
```diff
@@ -0,0 +1,34 @@
+dist: trusty
```
fstagni: Remove this file, all checks are done via Gitub Actions now.

Pilot/MessageSender.py
```diff
@@ -164,15 +171,22 @@ def sendMessage(self, msg, flag):
...
+      except (requests.RequestException,IOError) as e:
+        logging.error(e)
+        return False
+    else:
+      try:
+        requests.post(url, json=msg, cert=(hostCertificate, hostKey),
+                      verify=CACertificate)
+      except (requests.exceptions.RequestException,IOError) as e:
+        logging.error(e)
+        return False
```
fstagni: You can leave without this `if..else` block if in `backportRequests` you define `RequestException` as 

```
class exceptions(object):
  class RequestException(Exception):
    """Exception for requests"""

    def __init__(self, url, message):
      super(RequestException, self).__init__(url, message)
      self.url = url
      self.message = message

    def __str__(self):
      return '\nProblem with connecting to this url:\n{0}\nProblem:\n{1}'.format(self.url, self.message)
```

wkrzemien: Ok. Nice solution. I will use it

PR #106, opened 2020-02-28
Add JobAgent options: pollingTime, stopAppFailure, stopFailedMatches

aldbr:
This PR adds options related to `JobAgent` and bound to the `PoolCE` usage:
* `pollingTime`: mainly to reduce the polling time of `JobAgent` as it usually takes 30sec to submit a job to the `PoolCE`. It allows to start running multiple jobs quicker.
* `stopOnApplicationFailure`: boolean that can be set to `False` to avoid `JobAgent` to finish when an error occurs within a job. The `PoolCE` can run several jobs at the same time and we could continue the execution. Moreover, on batch systems, it could avoid to waste allocated time.
* `stopAfterFailedMatches`: same idea

It also removes the `MAX_CYCLES = 10` attribute limiting the number of cycles in `JobAgent` to 10.
As we may encounter some nodes having 24 or even 96 processors where we can run more than 10 jobs, we should be able to set `MaxCycles` without being limited.

### DIFF
--- a/Pilot/pilotCommands.py
+++ b/Pilot/pilotCommands.py
@@ -839,7 +839,10 @@ class LaunchAgent(CommandBase):
     self.inProcessOpts = ['-s /Resources/Computing/CEDefaults']
     self.inProcessOpts.append('-o WorkingDirectory=%s' % self.pp.workingDir)
     self.inProcessOpts.append('-o /LocalSite/CPUTime=%s' % (int(self.pp.jobCPUReq)))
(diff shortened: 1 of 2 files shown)


fstagni:
You can rebase this to master.

fstagni:
@atsareg please review this one too

fstagni:
@aldbr please rebase now, and then I will merge

aldbr:
@fstagni it's done

PR #111, opened 2020-06-10
Create dirac.cfg (more carefully)

sfayer:
Hi @fstagni,

Here is an updated version of the @marianne013 dirac.cfg patch with more code to check that the destination looks sensible. Would it be possible for you to try this with the LHCb set-up?

Regards,
Simon


marianne013:
I just tested Simon's patch in a more real life setting, and it seems to work.

sfayer:
Hi,
Is there any news on merging this?
Regards,
Simon

fstagni:
Sorry, I had forgotten about it. Merged now.

PR #115, opened 2020-07-23
Ensure pilot command output is always ascii

### LINKED ISSUE #112: Problem decoding output in pilotTools.py
Today I have tried to integrate a new resource on SunGridEngine. Job agent started, job status changed to Running. After that, it hangs, job on a resource is finished. In error file of the batch I see the following:
  File "dirac-pilot.py", line 66, in <module>
    command.execute()
  File "/tmp/DIRAC_6fagrxpilot/pilotCommands.py", line 917, in execute
    self.__startJobAgent()
  File "/tmp/DIRAC_6fagrxpilot/pilotCommands.py", line 904, in __startJobAgent
    retCode, _output = self.executeAndGetOutput(jobAgent, self.pp.installEnv)
  File "/tmp/DIRAC_6fagrxpilot/pilotTools.py", line 384, in executeAndGetOutput
    outData = _p.stdout.read().decode().strip()
UnicodeDecodeError: 'ascii' codec can't decode byte 0xd0 in position 5591: ordinal not in range(128)

chrisburr:
This PR ensures the return value of `executeAndGetOutput` is always valid ascii. Any invalid bytes are replaced with `?`. I've tested this will all sorts of odd locales and Python versions, including 2.6, 2.7 and 3.8. See the screenshot for examples.

In principle something similar could be done for `stderr` but it's only printed to `sys.stdout` so I don't think it's necessary.

Fixes #112
Closes #114 

Screenshot 2020-07-23 at 17 36 39

### DIFF
--- a/Pilot/pilotTools.py
+++ b/Pilot/pilotTools.py
@@ -381,10 +381,13 @@ class CommandBase(object):
                             stderr=subprocess.PIPE, close_fds=False)
 
       # standard output
-      outData = _p.stdout.read().decode().strip()
-      for line in outData:
-        sys.stdout.write(str(line))
-      sys.stdout.write('\n')
+      outData = _p.stdout.read()
+      # always convert the output to ascii
+      outData = outData.decode("ascii", "replace")
+      # replace any invalid characters with "?" to avoid having unicode output
+      outData = str(outData.replace(u"\ufffd", "?").strip())
+      # write to stdout for debugging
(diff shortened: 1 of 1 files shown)


fstagni:
Tested in https://jenkins-dirac.web.cern.ch/blue/organizations/jenkins/LHCbPilot3_CVM4_pipeline/detail/LHCbPilot3_CVM4_pipeline/94/pipeline/50 works fine.

PR #116, opened 2020-08-05
Scripts print more info, for debug purposes

fstagni:


Pilot/pilotCommands.py
```diff
@@ -781,37 +781,40 @@ def execute(self):
...
-    cpuNormalizationFactorOutput = cpuNormalizationFactorOutput.split('\n')[-1]
-
     # HS06 benchmark
-    # FIXME: this is a (necessary) hack!
-    cpuNormalizationFactor = float(cpuNormalizationFactorOutput.split('\n')[0].replace("Estimated CPU power is ",
-                                                                                       '').replace(" HS06", ''))
-    self.log.info(
-        "Current normalized CPU as determined by 'dirac-wms-cpu-normalization' is %f" %
-        cpuNormalizationFactor)
+    for line in cpuNormalizationFactorOutput.split('\n'):
```
chrisburr, line 790:
```suggestion
    cpuNormalizationFactor = None
    for line in cpuNormalizationFactorOutput.split('\n'):
```

Pilot/pilotCommands.py
```diff
@@ -781,37 +781,40 @@ def execute(self):
...
+      if "Estimated CPU power is" in line:
+        line = line.replace("Estimated CPU power is", '')
+      if "HS06" in line:
+        line = line.replace("HS06", '')
+        cpuNormalizationFactor = float(line.strip())
+        self.log.info(
+            "Current normalized CPU as determined by 'dirac-wms-cpu-normalization' is %f" %
+            cpuNormalizationFactor)
```
chrisburr, line 799: It's possible for `cpuNormalizationFactor` to be undefined here. Should it raise an error showing the command output so it's easier to debug?
```suggestion
    if cpuNormalizationFactor is None:
      raise NotImplementedError(cpuNormalizationFactorOutput)

```

(the same probably applies to `cpuTimeOutput` below)

fstagni: Yes I know. This applies also to `cpuTimeOutput` below. I thought about implementing what you suggested, but it should be, in theory, "impossible" so I prefer to leave it as it is.

PR #118, opened 2020-10-13
Change defaults for Pool CE

fstagni:


Pilot/pilotCommands.py
```diff
@@ -861,13 +861,21 @@ def __setInProcessOpts(self):
...
-    self.jobAgentOpts = ['-o MaxCycles=%s' % self.pp.maxCycles,
-                         '-o PollingTime=%s' % self.pp.pollingTime,
-                         '-o StopOnApplicationFailure=%s' % self.pp.stopOnApplicationFailure,
-                         '-o StopAfterFailedMatches=%s' % self.pp.stopAfterFailedMatches]
+    self.innerCEOpts = ['-s /Resources/Computing/CEDefaults']
+    self.innerCEOpts.append('-o WorkingDirectory=%s' % self.pp.workingDir)
+    self.innerCEOpts.append('-o /LocalSite/CPUTime=%s' % (int(self.pp.jobCPUReq)))
+    if self.pp.ceType == 'Pool':
+      self.jobAgentOpts = ['-o MaxCycles=%s' % max(self.pp.pilotProcessors, self.pp.maxCycles),
+                           '-o PollingTime=%s' % min(5, self.pp.maxCycles),
```
aldbr, line 869: Why `self.pp.maxCycles`? 
What about `min(5, self.pp.pollingTime)`?

Pilot/pilotCommands.py
```diff
@@ -861,13 +861,21 @@ def __setInProcessOpts(self):
...
+                           '-o StopOnApplicationFailure=%s' % True and self.pp.stopOnApplicationFailure,
```
aldbr, line 870: Why `True and self.pp.stopOnApplicationFailure`?
Isn't it the same thing to just have `self.pp.stopOnApplicationFailure`?

Pilot/pilotCommands.py
```diff
@@ -861,13 +861,21 @@ def __setInProcessOpts(self):
...
+                           '-o StopAfterFailedMatches=%s' % max(self.pp.pilotProcessors,
+                                                                self.pp.stopAfterFailedMatches),
+                           '-o FillingModeFlag=True']
```
aldbr, line 873: Wait, this was not present before? How did it work until now?

fstagni: It worked because, while the default in DIRAC is `False`, in LHCb configuration we override it (look for `FillingModeFlag` in the CS). This makes sure that this flag is always `True` for `PoolCE`

Pilot/pilotCommands.py
fstagni: oops, yes, you are right! I correct it.

Pilot/pilotCommands.py
fstagni: Of course yes, thanks! But I think that for PoolCE we should just put `False`...?

aldbr (approved):

Pilot/pilotCommands.py
aldbr: Yes I think so

Pilot/pilotCommands.py
aldbr: Oh yes alright, great

fstagni:
Tested in https://jenkins-dirac.web.cern.ch/view/DIRAC/job/Pilot3_submitAndMatch/104/console

Issue #113, opened 2020-07-15
Instrument the pilot to pass a GlobalDefaults.cfg file when installing DIRAC


fstagni:
As requested in https://github.com/DIRACGrid/DIRAC/issues/4604#issuecomment-630657095 instrument the pilot to pass a GlobalDefaults.cfg file when installing DIRAC

fstagni:
Won't solve this way, issues in https://github.com/DIRACGrid/management/issues

PR #138, opened 2021-09-21
[master] feat: added possibility to add a user-defined global defaults, from the pilot options

fstagni:
solves #113 

@zhangxiaomei Have a look, IIUC this is what you want

- [ ] port it to `management` repo if approved.

### DIFF
--- a/Pilot/dirac-install.py
+++ b/Pilot/dirac-install.py
@@ -357,10 +357,7 @@ class ReleaseConfig(object):
     :param str projectName: the name of the project
     :param str globalDefaultsURL: the default url
     """
-    if globalDefaultsURL:
-      self.globalDefaultsURL = globalDefaultsURL
-    else:
-      self.globalDefaultsURL = "http://diracproject.web.cern.ch/diracproject/configs/globalDefaults.cfg"
+    self.user_globalDefaultsURL = globalDefaultsURL
     self.globalDefaults = ReleaseConfig.CFG()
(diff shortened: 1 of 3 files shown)


zhangxiaomei:
Looks fine to us, thanks.

fstagni:
@atsareg review?

zhangxiaomei:
Look at again and have one question:
I haven't seen user_globalDefaultsURL was used here.  The logic I saw is: first look at globalDefaultsCVMFSPath -> If not available, go to use diracproject.web.cern.ch -> If not available, return

atsareg:
I do not see where global defaults are loaded from self.user_globalDefaultsURL in dirac-install.py. Wasn't it forgotten ?

fstagni:
> Look at again and have one question:
> I haven't seen user_globalDefaultsURL was used here. The logic I saw is: first look at globalDefaultsCVMFSPath -> If not available, go to use diracproject.web.cern.ch -> If not available, return

`user_globalDefaultsURL` is used, when requested, instead of `globalDefaultsCVMFSPath`. If `user_globalDefaultsURL` is not defined, and `globalDefaultsCVMFSPath` is not reachable, it will use diracproject.web.cern.ch. 

> I do not see where global defaults are loaded from self.user_globalDefaultsURL in dirac-install.py. Wasn't it forgotten ?

Ah right. It should be better now.

zhangxiaomei (approved):

PR #144, opened 2021-12-07
[devel] Pip install from URI

fstagni:
This PR adds the possibility to run `pip install DIRAC` from a generic URI. The mechanism used is simply an extension of the "modules" one, meaning we can do (as before) `pip install git+https://github.com/fstagni/DIRAC.git@mybranch#egg=DIRAC[pilot]` or, for example `pip install /some/path/[pilot]` which basically adds support for pip installing a package that is on CVMFS.


fstagni:
review, please!

PR #145, opened 2021-12-17
fix: follow https://github.com/DIRACGrid/management/pull/50

fstagni:
@iueda @hmiyake I think this is what you mentioned some time ago... (and I forgot until now).

Please review.


iueda:
Yes, exactly.

Since then I was also wondering if the lines 1609-1612 and 1622-1625 can stay as they are.
We have not faced any issue with X509_CERT_DIR and SSL_CERT_DIR, so leaving them is ok, but I am wondering if we can trust /etc/grid-security/certificates even when /etc/grid-security/vomsdir cannot be...?

iueda:
I think this can go in as is. My additional comment can be discussed separately.

PR #149, opened 2022-03-08
redirect stderr of DIRACOS2 to stdout, remove diracos dir if already existing

fstagni:


Pilot/pilotCommands.py
```diff
@@ -263,8 +264,11 @@ def _installDIRACpy3(self):
 
         # 2. Try to install from CVMFS
 
+        if os.path.exists("diracos"):
+            shutil.rmtree("diracos")
+
         retCode, _ = self.executeAndGetOutput(
-            "bash /cvmfs/dirac.egi.eu/installSource/%s" % installerName, self.pp.installEnv
+            "bash /cvmfs/dirac.egi.eu/installSource/%s 2>&1" % installerName, self.pp.installEnv
```
fstagni, line 271: Here I am wondering if we should better have
```suggestion
            "bash /cvmfs/dirac.egi.eu/installSource/%s 1>/dev/null 2>&1" % installerName, self.pp.installEnv
```
or 
```suggestion
            "bash /cvmfs/dirac.egi.eu/installSource/%s 1>%s.out 2>&1" % installerName, installerName, self.pp.installEnv
```

chrisburr: I think I would be inclined towards `2>&1`.

chrisburr (approved):

PR #150, opened 2022-03-10
feat: support multi-core allocations to compute cpu power

aldbr:
Need this PR to be merged before: https://github.com/DIRACGrid/DIRAC/pull/5939

Passing the number of processors available in the allocation to `dirac-wms-get-cpu-normalization`.

### DIFF
--- a/Pilot/pilotCommands.py
+++ b/Pilot/pilotCommands.py
@@ -893,11 +893,24 @@ class ConfigureCPURequirements(CommandBase):
             else:
                 configFileArg = "%s -R %s %s" % (configFileArg, self.pp.localConfigFile, self.pp.localConfigFile)
         retCode, cpuNormalizationFactorOutput = self.executeAndGetOutput(
-            "dirac-wms-cpu-normalization -U %s -d" % configFileArg, self.pp.installEnv
+            "dirac-wms-cpu-normalization --NumberOfProcessors=%d -U %s -d" % (self.pp.pilotProcessors, configFileArg),
+            self.pp.installEnv
         )
         if retCode:
-            self.log.error("Failed to determine cpu normalization [ERROR %d]" % retCode)
-            self.exitWithError(retCode)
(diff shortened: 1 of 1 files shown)


fstagni:
If we merge this PR before the DIRAC one, we'll get into

```
# dirac-wms-cpu-normalization --NumberOfProcessors=3
Error when parsing command line arguments: option --NumberOfProcessors not recognized
```
and if we merge it after, we'll get anyway in this same situation because of users/installations installing older versions of DIRAC.

fstagni:
It is MAYBE possible to extract the script from DIRAC and add it here, but I think it is more realistic a (ugly) "try with `--NumberOfProcessors` option and remove it if it fails"

aldbr:
> and if we merge it after, we'll get anyway in this same situation because of users/installations installing older versions of DIRAC.

Oops yes indeed...

> It is MAYBE possible to extract the script from DIRAC and add it here, but I think it is more realistic a (ugly) "try with `--NumberOfProcessors` option and remove it if it fails"

Yes, the second solution seems easier but ugly...

What if:
- I remove the `NumberOfProcessors` argument from `dirac-wms-cpu-normalization`
- `dirac-wms-cpu-normalization` directly gets `numberOfProcessors` from the configuration that we pass?

fstagni:
> What if:
> 
>     * I remove the `NumberOfProcessors` argument from `dirac-wms-cpu-normalization`
> 
>     * `dirac-wms-cpu-normalization` directly gets `numberOfProcessors` from the configuration that we pass?


Looks reasonable. You can get it from `/Resources/Computing/CEDefaults/NumberOfProcessors`

aldbr:
Closing this PR.
The following PR should be enough to support multi-core DB12: https://github.com/DIRACGrid/DIRAC/pull/5939
It tries to get the number of processors from the CLI first, and then from the `cfg` file.

PR #152, opened 2022-03-16
feat: Enable remote pilot logging using REST interface to Tornado.

martynia:
This is a revived remote pilot logger using REST interface to Tornado. No other interfaces (local file, message queue) were modified.


fstagni:
There are conflicts. In any case, please target `devel` branch.

PR #169, opened 2022-11-14
Jenkins fixes

fstagni:


fstagni:
`devel` branch and in any case I don't think it really needs a review...

PR #176, opened 2022-12-13
[devel] Remote Pilot Logger to Tornado new JSON format

martynia:
Modified pilot code to accept a new pilot.json format which closer follows CS Operations section.

The new  pilot code is backward compatible with the old format. A format selection is made on a `Setups` key which only exists in the old format. 

`PilotCStoJSONSynchronizer` dumps the entire `Operations` section to _JSON._ After reading in and parsing the _JSON_ file,  innermost dictionaries are scanned for options in a specific order:

```python
       paths = [
            "/Defaults/Pilot",
            "/%s/Pilot" % self.setup,
            "/%s/Defaults/Pilot" % vo,
            "/%s/%s/Pilot" % (vo, self.setup),
        ]
``` 

which greatly simplifies the original code, where an endless collection of  `if-else` statements was difficult to follow. The paths searched may be extended, if needed.

Various pilot options (like logging-related ones), `Commands` and `CommandExtensions` are searched this way. The CE description code is taken from the old code w/o modifications. Other parameters, like `Version` and `ConfigurationServers` are obtained in a similar way as before.


Pilot/proxyTools.py
```diff
@@ -0,0 +1,61 @@
...
+    return out.split(b"\n")
+
+
+def findExtension(oid, lines):
+    for i, line in enumerate(lines):
+        if oid in line:
+            return i
+
+
+def getVO(proxy_data):
```
chrisburr, line 25: This function should have unit tests assoicated with it (preferably with fake proxies of a few different levels, `voms-proxy-fake` should be able to help generate them).

martynia:
Hi,

Thanks for your comments.

Thanks for a fake proxy hint, I was wondering how I could put a real proxy in a test…

Now, I tested the code locally, and managed (sort of… see below) to get it working in Python 2 and 3.

But:

re.compile(br"^\s*\d+:d=(\d+)\s+hl=“)
Gives a syntax error, the line below doesn’t:
re.compile(rb"^\s*\d+:d=(\d+)\s+hl=“)

On my test machine with Python 2.75

I submitted a job and got this:

 File "/srv/localstage/condor/dir_24554/DIRAC_7EernFpilot/proxyTools.py", line 9
   RE_OPENSSL_ANS1_FORMAT = re.compile(rb"^\s*\d+:d=(\d+)\s+hl=")
                                                               ^
SyntaxError: invalid syntax
This is the same error when trying the code with (br”…”) above. Binary string should do no harm in Linux, so what is a problem with this ?
Thanks, JM

chrisburr:
I think you're seing a mixture of issues here:

* In Python 2 the only valid prefix is `br"` (python 3 allows any order): https://docs.python.org/2.7/reference/lexical_analysis.html#string-literals
* The code you've pasted contains a "Left Double Quotation Mark" rather than a regular double quote

martynia:
The quotes got mixed up when pasting, they are OK in the code.

My local  black  changed br to rb when saving a file and I pushed a broken code !

OK, it should be better now.

Pilot/proxyTools.py
martynia: I have written a simple test to see if a `voms-proxy-fake` creates a proxy and tested it a an ubuntu container. 
I would also need a set of fake certificates with are in Dirac here: `/DIRAC/src/DIRAC/Core/Security/test/certs`
A change which I had to make was to install voms with `conda install -c conda-forge voms`, which would go right after the line:
[https://github.com/DIRACGrid/Pilot/blob/3a837eccbf9f494d16e504699db1e32c3dc683d5/.github/workflows/basic.yml#L31
](https://github.com/DIRACGrid/Pilot/blob/3a837eccbf9f494d16e504699db1e32c3dc683d5/.github/workflows/basic.yml#L31)

However there seem to be an issue with running `pytest.` The line:
[https://github.com/DIRACGrid/Pilot/blob/3a837eccbf9f494d16e504699db1e32c3dc683d5/.github/workflows/basic.yml#L41
](https://github.com/DIRACGrid/Pilot/blob/3a837eccbf9f494d16e504699db1e32c3dc683d5/.github/workflows/basic.yml#L41)

prevents `pytest` from being run, doesn't it ?

fstagni:
Please rebase now (you also have conflicts)

martynia:
@fstagni, could you please verify why a `shellcheck` is failing (undefined variable) ?

chrisburr:
> Note: This message only triggers for variables with lowercase characters in their name (foo and kFOO but not FOO) due to the standard convention of using lowercase variable names for unexported, local variables.

If it's supposed to be an environment variable that is set elsewhere it should be all caps.

fstagni (approved):

fstagni:
This PR will now be merged (in devel branch only, of course) so it can be tested in tomorrow's hackathon.

PR #177, opened 2023-01-20
[devel] fully removed dirac-install and python2 DIRAC client installations

fstagni:


fstagni:
Un-mergeable, will re-do.

PR #179, opened 2023-02-03
[master] Completely rewrote GitHub Actions workflow

### LINKED ISSUE #178: Logic in basic.yml prevents pytest from being run
There is a weird logic in this script. The line:
[https://github.com/DIRACGrid/Pilot/blob/3a837eccbf9f494d16e504699db1e32c3dc683d5/.github/workflows/basic.yml#L41
](https://github.com/DIRACGrid/Pilot/blob/3a837eccbf9f494d16e504699db1e32c3dc683d5/.github/workflows/basic.yml#L41)

will prevent `pytest` from being ever run. Not sure what was a reason behind this.

fstagni:
closes #178


martynia:
I cherry-picked those changes to my `devel`-derived branch. A problem is, that the` if` statement: 
`if: github.event_name != 'push' || github.repository == 'DIRACGrid/Pilot'`
might be too restrictive - no test will run on user pilot repositories on push. Is it intentional ?

fstagni:
> I cherry-picked those changes to my `devel`-derived branch. A problem is, that the` if` statement: `if: github.event_name != 'push' || github.repository == 'DIRACGrid/Pilot'` might be too restrictive - no test will run on user pilot repositories on push. Is it intentional ?

It is easy enough to reproduce all these tests on the local development environment that I don't see the need to run these tests also on the forks.

chrisburr (approved):

PR #183, opened 2023-03-28
fix: update worker

fstagni:


fstagni:
Merging, this is only for the Jenkins tests...

PR #184, opened 2023-03-29
[devel] Vacuum pilots

fstagni:
This is the same as https://github.com/DIRACGrid/Pilot/pull/181 just targeting `devel` branch instead of `master`


fstagni:
Review OK?

Issue #166, opened 2022-11-03
Support taking unpacked install from CVMFS


chrisburr:
It would be nice for the Pilot to take an existing DIRACOS installation from CVMFS to reduce the load on the worker node's filesystem.

I've put an example on `/cvmfs/lhcbdev.cern.ch` which would probably be hosted on `/cvmfs/dirac.egi.eu` instead (Don't use it, I may delete it at any time!). The idea would then be to make a Python `venv` on top of it like so:

```bash
$ source /cvmfs/lhcbdev.cern.ch/experimental/dirac-pilot-ideas/v1/DIRACOS/v2.29/Linux-x86_64/diracosrc
$ python -m venv --system-site-packages /tmp/diracos-venv
$ cp $DIRACOS/diracosrc /tmp/diracos-venv/diracosrc
$ echo "source /tmp/diracos-venv/bin/activate" >> /tmp/diracos-venv/diracosrc
# Some other modifications should probably also be done to the diracosrc
# From now onwards "source /tmp/diracos-venv/diracosrc" can be used to activate the CVMFS based DIRACOS
$ source /tmp/diracos-venv/diracosrc
$ pip install 'DIRAC==7.3.32'
```

This avoids creating 62108 files/links/directories (1.6GB of data) and would make pilots much faster while still letting people control the DIRAC version and install extensions if desired.

My suggestion to implement this would be:

* [ ] Add a script to DIRACOS which creates this virtual environment
* [ ] Add support to the pilot to do this instead of installing DIRACOS2 (might also need a little SiteDirector work)
* [ ] Setup automatic installation of versions DIRACOS environments on /cvmfs/dirac.egi.eu

cc @IgorPelevanyuk who expressed interest in this before the BiLD meeting

IgorPelevanyuk:
Great! Thank you!
I will try to look into this.

fstagni:
PR https://github.com/DIRACGrid/Pilot/pull/205 is for this. There are anyway few points:

- If we would say that the policy is to have all releases on CVMFS, then, by default, we would not need to set any of the `--preinstalledEnv` flags
- the mechanism could be valid also for extensions, if such would have the possibility to deploy on `cvmfs.dirac.egi.eu`
- In general, I think that releases should be deployed as soon as they are done (we had something related to this in https://github.com/DIRACGrid/management/issues/11 already few years ago)

fstagni:
Achieved. For the moment in `devel` branch.

PR #187, opened 2023-05-26
[master] Setting up preinstalled DIRAC in a pilot

atsareg:
This PR allows to specify the setup script location which will be used to define a preinstalled DIRAC environment instead of installing DIRACOS2/DIRAC from scratch on a worker node. Example script locations:

   - /cvmfs/dirac.egi.eu/dirac/pro/diracos/diracosrc
   - /cvmfs/dirac.egi.eu/dirac/bashrc_egi

  This is addressing #166 but makes use of predefined DIRAC clients in CVMFS.


fstagni:
This is a bit different/simpler solution to #166 , the question here if this would be enough. 
Using a `venv`, as suggested in the original issue, would add flexibility on top of what is here.

chrisburr:
This effectively bypasses the pilot version specified in the CS which doesn't feel great. It also doesn't work if you have multiple architectures. If we could instead have a layout like:

```bash
/cvmfs/dirac.egi.eu/dirac/{version}/{uname -m}-{uname -s}/diracosrc
```

You can then have `--preinstalledEnvPrefix=/cvmfs/dirac.egi.eu/dirac/` and would probably be good enough to replace the `LHCbInstallDIRAC` pilot command.

atsareg:
In CVMFS we have now /cvmfs/dirac.egi.eu/dirac/{version}/diracos/diracosrc. So a particular DIRAC version can be specified. But not different platforms. In LHCb you have different platforms installed in cvmfs ?

chrisburr:
I'd propose having `--preinstalledEnvBashrc=` in addition to `--preinstalledEnvPrefix=/cvmfs/dirac.egi.eu/dirac/` with the view that it contains at least something like:

```
source ${DIRACOS}/diracosrc
```

atsareg:
The PR is updated along the lines of the discussion above. I keep to settings, --preinstalledEnv and --preinstalledEnvPrefix. If --preinstalledEnv is specified, it is used as the environment setting script, --preinstalledEnvPrefix is ignored in that case. If --preinstalledEnvPrefix is specified  (and not  --preinstalledEnv), then environment is resolved with the release area structure like in the CVMFS of LHCb. I have updated the /cvmfs/dirac.egi.eu to this structure already

fstagni:
One more note: the case of user-defined environments is not considered in this implementation.

fstagni:
Closing as superseded by #205

PR #190, opened 2023-05-31
Limit Mamba concurrency during installation

### LINKED ISSUE #189: Setup of DIRACOS does not restrict cpu cores
When the Pilot creates its DIRACOS environment, it [directly calls the `DIRACOS-Linux-machine.sh`](https://github.com/DIRACGrid/Pilot/blob/63ac77b85bcf3582f48dd4b3f81592bf33cda6be/Pilot/pilotCommands.py#L307-L309) which eventually invokes `mamba`. Mamba assumes that it can use *all* cores of the machine (see https://github.com/mamba-org/mamba/issues/2463), which isn't realistic for Pilot environments. This leads to excessive process creation, which can negatively affect the pilot, user or even entire compute resource.

As far as I can tell, the templates from which DIRACOS is generated do no provide a feasible way to limit this internally. The Pilot thus seems like the best place, seeing how it is aware of resource restrictions.  
A solution would be to set `MAMBA_EXTRACT_THREADS` when installing DIRACOS, either to a conservative `1` or `pp.maxNumberOfProcessors`.

---------------

For reference of scale, we caught this on a WLCG Tier 1 WN with 256 cores that got allocated mostly to one VO. Each of the *single* core pilots tried to use *256* child processes; each pilot quickly ground to a halt due to resource and fork bomb protection, which caused each new pilot to also immediately get stuck on nproc limits and similar safeguards.

maxfischer2781:
This PR limits the resources used during installation to what the pilot actually owns. It uses the `maxNumberOfProcessors` with a default of `1`; both `self.pp.payloadProcessors` and `self.pp.pilotProcessors` are not available during default command order.

Closes #189.

### DIFF
--- a/Pilot/pilotCommands.py
+++ b/Pilot/pilotCommands.py
@@ -289,6 +289,12 @@ class InstallDIRAC(CommandBase):
 
     def _installDIRACpy3(self):
         """Install python3 version of DIRAC client"""
+        # default to limit the resources used during installation to what the pilot owns
+        installEnv = {
+            # see https://github.com/DIRACGrid/Pilot/issues/189
+            "MAMBA_EXTRACT_THREADS": self.pp.maxNumberOfProcessors or 1,
+        }
+        installEnv.update(self.pp.installEnv)
 
         # 1. Get the DIRACOS installer name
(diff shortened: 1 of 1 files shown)


fstagni:
I don't see anything wrong with this, but the mamba expert here is @chrisburr . Any suggestion?

chrisburr (approved):

PR #192, opened 2023-06-20
fix: adding DIRACSYSCONFIG (for pilot.cfg) to local environment

fstagni:


sfayer:
I can't make a review suggestion, but here is the diff for my tweaked version:
```diff
-            os.environ["DIRACSYSCONFIG"] = os.path.realpath(self.pp.localConfigFile)
+            self.pp.installEnv["DIRACSYSCONFIG"] = os.path.realpath(self.pp.localConfigFile)
```

PR #196, opened 2023-06-28
[master] add new command RegisterPilot (second attempt)

fstagni:
Merging https://github.com/DIRACGrid/Pilot/pull/181 did not go well, and was reverted in https://github.com/DIRACGrid/Pilot/pull/195. 

This is a second attempt at making this right!


chrisburr:
It it worth running with this branch in lhcb prod for a few days to see that it actually seems to work without affecting other installations?

fstagni:
> It it worth running with this branch in lhcb prod for a few days to see that it actually seems to work without affecting other installations?

After that attempt, we ran in https://lblogbook.cern.ch/Operations/36604. Which, IIUC, it is now fixed, but @aldbr should give a second opinion. 

We can also re-do the exercise above.

aldbr (approved):
> After that attempt, we ran in https://lblogbook.cern.ch/Operations/36604. Which, IIUC, it is now fixed, but @aldbr should give a second opinion.

The `GRID_GLOBAL_JOBURL` env variable related to ARC is used, so it should be fine now.

fstagni:
OK. Testing it now in LHCb production.

Issue #197, opened 2023-07-03
Question: Tags/RequiredTags added twice to `pilot.cfg`


aldbr:
Not a real issue, just a question related to https://github.com/DIRACGrid/DIRAC/issues/7086

`Tags/RequiredTags` are added both in `/LocalSite` and `/Resources/Computing/CEDefaults`. Is that expected?

See here:
https://github.com/DIRACGrid/Pilot/blob/15a0874e520a83243aee2fdcec20a87ccc1328d6/Pilot/pilotCommands.py#L565-L568

And here:
https://github.com/DIRACGrid/Pilot/blob/15a0874e520a83243aee2fdcec20a87ccc1328d6/Pilot/pilotCommands.py#L679-L687

Thanks

fstagni:
Looks to me that they are only added in 

https://github.com/DIRACGrid/Pilot/blob/15a0874e520a83243aee2fdcec20a87ccc1328d6/Pilot/pilotCommands.py#L679-L687
 

??

aldbr:
`self.pp.tags` is created from `resourceDict` here:

https://github.com/DIRACGrid/Pilot/blob/15a0874e520a83243aee2fdcec20a87ccc1328d6/Pilot/pilotCommands.py#L576

And `self.pp.queueParameters` is also created from `resourceDict` and then used to feed `self.cfg` with `/LocalSite` options:

https://github.com/DIRACGrid/Pilot/blob/15a0874e520a83243aee2fdcec20a87ccc1328d6/Pilot/pilotCommands.py#L564-L568

Am I wrong?

fstagni:
Bof, anyway the 2 locations are rather equivalent, so leave 1 only (none of them is obviously better than the other)

PR #198, opened 2023-07-10
devel: pass tags to the Pilot config only once

aldbr:
Aims at "solving" https://github.com/DIRACGrid/Pilot/issues/197

Currently:
- `tags`/`reqtags` are present in two different locations in the configuration: 
  - `/LocalSite`: can contain many times the same tags
  - `/Resources/Computing/CEDefaults`: tags are "filtered" such as duplicates are removed

In this PR:
- `tags`/`reqtags` are only present in `/Resources/Computing/CEDefaults`

### DIFF
--- a/Pilot/pilotCommands.py
+++ b/Pilot/pilotCommands.py
@@ -740,23 +740,21 @@ class CheckCECapabilities(CommandBase):
             self.log.error("The pilot command output is not json compatible.")
             self.exitWithError(1)
 
-        self.pp.queueParameters = resourceDict
-        for queueParamName, queueParamValue in self.pp.queueParameters.items():
-            if isinstance(queueParamValue, list):  # for the tags
-                queueParamValue = ",".join([str(qpv).strip() for qpv in queueParamValue])
-            self.cfg.append("-o /LocalSite/%s=%s" % (queueParamName, quote(queueParamValue)))
-
         # Pick up all the relevant resource parameters that will be used in the job matching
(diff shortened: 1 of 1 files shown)


Pilot/pilotCommands.py
```diff
@@ -513,23 +513,21 @@ def execute(self):
...
         # RequiredTags are like Tags.
-        if resourceDict.get("RequiredTag"):
-            self.pp.reqtags += resourceDict["RequiredTag"]
+        self.pp.reqtags += resourceDict.pop("RequiredTag", [])
+
+        self.pp.queueParameters = resourceDict
+        for queueParamName, queueParamValue in self.pp.queueParameters.items():
+            if isinstance(queueParamValue, list):  # for the tags
+                queueParamValue = ",".join([str(qpv).strip() for qpv in queueParamValue])
+            self.cfg.append("-o /LocalSite/%s=%s" % (queueParamName, quote(queueParamValue)))
```
fstagni, line 530: In the description of the PR you wrote:

```
`tags/reqtags` are only present in `/Resources/Computing/CEDefaults`
```
but here you are adding them to `/LocalSite`?

aldbr: I add them in `self.pp.tags/reqtags` and pop them from `resourceDict` before `self.pp.queueParameters = resourceDict`, so they should not be present in `/LocalSite`, am I doing it wrong?

fstagni (approved):

PR #199, opened 2023-07-10
remove: old variables from setFlavour()

aldbr:
- `CREAM` and `GLOBUS` are not used anymore
- `OSG`: what is this?
- `ARC`: `GRID_GLOBAL_JOBURL` is always present


andresailer:
OSG is the OpenScienceGrid. The grid in the US mostly. They should be all HTCondorCE by now, so this can be removed, most likely.

fstagni (approved):

PR #202, opened 2023-08-01
[devel] add new command RegisterPilot (second attempt)

fstagni:
Same as https://github.com/DIRACGrid/Pilot/pull/196 but for `devel` branch


fstagni:
Merging, as this is for `devel` branch, in order to also test in certification setup.

PR #205, opened 2023-08-01
[devel] Setting up preinstalled DIRAC in a pilot

fstagni:
This is in effect a replacement of https://github.com/DIRACGrid/Pilot/pull/187

Tested in https://jenkins-dirac.web.cern.ch/view/DIRAC/job/Pilot3_CVM4_pipeline/450, https://jenkins-dirac.web.cern.ch/view/DIRAC/job/Pilot3_CVM4_pipeline/451
+ in https://gitlab.cern.ch/lhcb-dirac/LHCbPilot/-/jobs/31453772 for LHCb compatibility.


fstagni:
OK to merge? This can be tested in tomorrow's hackathon.

Pilot/pilotCommands.py
```diff
@@ -431,16 +451,58 @@ def _installDIRACpy3(self):
...
+                    if self.pp.userEnvVariables:
+                        # User-requested environment variables (comma-separated, name and value separated by ":::")
+                        # The case of dirac-install is handled in vanilla DIRAC Pilot
+                        newEnvVars = dict(
+                            name.split(":::", 1) for name in self.pp.userEnvVariables.replace(" ", "").split(",")
+                        )
+                        self.log.info("Adding env variable(s) to the environment", newEnvVars)
+                        self.pp.installEnv.update(newEnvVars)
+                else: 
+                    self._localInstallDIRAC()
```
andresailer, line 489:
```suggestion
            # In case we want to force local installation (in absence of CVMFS or for test reasons)
            if "diracInstallOnly" in self.pp.genericOption:
                self.log.info("NOT sourcing: starting traditional DIRAC installation")
                self._localInstallDIRAC()
                return
            
            # Try sourcing from CVMFS
            self._getPreinstalledEnvScript()
            if not self.pp.preinstalledEnv:
                self._localInstallDIRAC()
                return

            # if we are here, we have a preinstalled environment
            self._sourceEnvironmentFile()
            self.log.info("source DIRAC env DONE, for release %s" % self.pp.releaseVersion)
            # environment variables to add?
            if self.pp.userEnvVariables:
                # User-requested environment variables (comma-separated, name and value separated by ":::")
                # The case of dirac-install is handled in vanilla DIRAC Pilot
                newEnvVars = dict(
                name.split(":::", 1) for name in self.pp.userEnvVariables.replace(" ", "").split(",")
              )
                self.log.info("Adding env variable(s) to the environment", newEnvVars)
                self.pp.installEnv.update(newEnvVars)
```

Even with the `return` the `finally` block gets executed, so this gives a flatter code-path, which I find easier to reason about.

fstagni: I think the `else` is necessary here, otherwise we might end up installing twice.

andresailer: Which `else`? Did I miss a `return` or something else?

andresailer: PS: I could have also messed up the indentation, because I did that in the github comment window.

Pilot/pilotTools.py
```diff
@@ -180,6 +187,24 @@ def which(cmd):
     return False
 
 
+def safe_listdir(directory, timeout=60):
```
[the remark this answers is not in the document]
fstagni: Done.

Pilot/pilotCommands.py
fstagni: Maybe I figured it out, check the version I just pushed.

Pilot/pilotTools.py
```diff
@@ -180,6 +187,32 @@ def which(cmd):
...
+    """ This is "safe" list directory,
+    for lazily-loaded File Systems like CVMFS.
+    There's by default a 60 seconds timeout.
+
+    :param str directory: directort to list
```
andresailer, line 195:
```suggestion
    :param str directory: directory to list
```

Pilot/pilotCommands.py
```diff
@@ -298,7 +301,24 @@ def _locateInstallationScript(self):
...
+        """ Get preinstalled environment script if any """
+
+        preinstalledEnvScript = self.pp.preinstalledEnv
+        if not preinstalledEnvScript and self.pp.preinstalledEnvPrefix:
+            version = self.pp.releaseVersion or "pro"
+            arch = platform.system() + "-" + platform.machine()
+            preinstalledEnvScript = os.path.join(self.pp.preinstalledEnvPrefix, version, arch, "diracosrc")
+
+        self.log.info("Evaluating env script %s" % preinstalledEnvScript)
+        if not safe_listdir(preinstalledEnvScript):
```
andresailer, line 314: This is calling `safe_listdir` on a file. Which results in "Not a directory"
```
touch aFile; python -c "import os; os.listdir('aFile')"
  File "<string>", line 1, in <module>
OSError: [Errno 20] Not a directory: 'aFile'
```

andresailer: PS: with python 3.9
```
touch aFile; python -c "import os; os.listdir('aFile')"
  File "<string>", line 1, in <module>
NotADirectoryError: [Errno 20] Not a directory: 'aFile'
```

fstagni: Ouf, thanks. Corrected.

fstagni:
After giving it some thoughts, I think that this one for the moment can stay as it is, even the the caveats above.

atsareg:
I think this can be merged. As for the releases - I try to keep them up to date in the CVMFS. In any case the pro link is defined, so in most cases, It would be enough to define only the preinstalledEnvPrefix.

atsareg (approved):

PR #207, opened 2023-09-21
[devel] add a pathelement to be searched for in pilot.json

martynia:
Add a `/<voname>/Pilot` path element to be searched  for when pilot.json is parsed.

### DIFF
--- a/Pilot/pilotTools.py
+++ b/Pilot/pilotTools.py
@@ -1114,6 +1114,7 @@ class PilotParams(object):
         vo = self.__getVO()
         paths = [
             "/Defaults/Pilot",
+            "/%s/Pilot" % vo,
             "/%s/Pilot" % self.setup,
             "/%s/Defaults/Pilot" % vo,
             "/%s/%s/Pilot" % (vo, self.setup),


Pilot/pilotTools.py
```diff
@@ -1114,9 +1114,8 @@ def __getSearchPaths(self):
         vo = self.__getVO()
         paths = [
             "/Defaults/Pilot",
-            "/%s/Pilot" % self.setup,
+            "/%s/Pilot" % vo,
             "/%s/Defaults/Pilot" % vo,
-            "/%s/%s/Pilot" % (vo, self.setup),
         ]
```
fstagni, line 1119: We need to be compatible also with version of the CS that still have self.setup in it.

martynia: OK, last commit removed.

marianne013:
@fstagni Janusz is quite keen to finish this, could this be merged ? Incidentally do you have any idea when we can expect running jobs on lbcertificate70 ?

fstagni (approved):

PR #208, opened 2023-10-02
[devel] Better reporting of errors

fstagni:


martynia (commented):
It looks fine to me. As long as `self.exitWithError()` calls exit() it fine (the log finaliser runs on exit)

fstagni:
> It looks fine to me. As long as `self.exitWithError()` calls exit() it fine (the log finaliser runs on exit)

Yes it does.

PR #209, opened 2023-10-02
[devel] feat: removed check for --cfg

fstagni:
As, I believe, everyone's using at least version v7r0...


martynia (commented):
Fine with me.

PR #210, opened 2023-10-02
test: adding integration tests

fstagni:
This PR adds few integration test for the pilot:

1. The test simply installs the Pilot, getting configuration from lbcertifdirac70. The matrix variables are right now the python version (the version found on the WN) and the release version, for which here we take the `HEAD` of `rel-v8r0` and `integration` branches of DIRAC. More variables to this matrix can be added, if felt requested.
2. A test where DIRAC is sourced from CVMFS
3. A test for pilot that is installed and then tries to match a job (from lbcertifdirac70)
4. A test where DIRAC is sourced from CVMFS and tries to match a job
5. A integration test for the LHCb extension
6. A integration test for the LHCb extension, but using the vanilla Pilot DIRAC installer

These tests are only run on "push", so not in the pull requests (for a reason that I ignore, the base64 encoded certs used here are not properly passed in the environment of github actions runners for PRs). The tests ran on my fork: https://github.com/fstagni/Pilot/actions/workflows/integration.yml (after adding the correct secrets...).

The tests for LHCb are just an example. If you have your own Pilot extension, those specific tests can be added here: cc @sfayer @hmiyake @arrabito

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -163,7 +163,7 @@ jobs:
           sed -i "s/VAR_JENKINS_SITE/DIRAC.Jenkins.ch/g" pilot.json
           sed -i "s/VAR_JENKINS_CE/jenkins-lhcb.cern.ch/g" pilot.json
           sed -i "s/VAR_JENKINS_QUEUE/jenkins-queue_not_important/g" pilot.json
-          sed -i "s/VAR_DIRAC_VERSION//g" pilot.json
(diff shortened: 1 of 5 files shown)


arrabito:
Thanks for these tests. As CTA for now we don't have any pilot extension.

fstagni:
> Thanks for these tests. As CTA for now we don't have any pilot extension.

Thanks for the info. What is the full command to start CTA pilots? To see if these tests cover you case.

fstagni:
I merge this, I see nothing wrong with it. No review, as usual for my PRs, can't wait forever.

PR #214, opened 2023-10-12
[devel] Fixed format of json file in test, added to integration tests matrix

fstagni:


fstagni:
Integration tests running (and passing, I think) in https://github.com/fstagni/Pilot/actions/runs/6497343580

PR #217, opened 2023-10-27
[devel] feat: add log flushing with a timer.

martynia:
Added a timer which runs in a separate thread and flushed a remote logger buffer periodically. The timer can be disabled and in this case the buffer will be flushed when full as before.


fstagni:
Ran also integration tests in https://github.com/fstagni/Pilot/actions/runs/6708184649

fstagni (approved):

PR #221, opened 2023-11-07
[devel] fix RemoteLogger activation logic; pilot logger timer fixes

martynia:
Fix the logic when a remote logger is activated. There are 2 options in the CS:

- RemoteLoggerURL
- RemoteLogging flag

The remote logger is activated when the URL` is not None` and the flag is set to `True`.
The logic before was kinda broken, a disabled remote logger would not emit remote messages(OK) and it would act as a classic logger (OK) but caused problems when a timer was introduced and was not correctly canceled.

### DIFF
--- a/Pilot/tests/Test_simplePilotLogger.py
+++ b/Pilot/tests/Test_simplePilotLogger.py
@@ -10,9 +10,9 @@ import sys
 import tempfile
 
 try:
-    from Pilot.pilotTools import CommandBase, PilotParams, RemoteLogger
+    from Pilot.pilotTools import CommandBase, PilotParams, Logger, RemoteLogger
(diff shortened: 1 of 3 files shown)


Pilot/tests/Test_simplePilotLogger.py
```diff
@@ -122,14 +122,17 @@ def tearDown(self):
...
             "-d",
             "-g",
             "dummyURL",
             "-F",
             "tests/pilot.json",
         ]
-
+        import os
+        cwd = os.getcwd()
+        print("cwd", cwd)
```
fstagni, line 135: remove?

fstagni (approved):

PR #222, opened 2023-11-10
[devel] Collect batch system info

aldbr:
Aims at solving https://github.com/DIRACGrid/DIRAC/issues/7177
Associated with https://github.com/DIRACGrid/DIRAC/pull/7289

### DIFF
--- a/Pilot/pilotTools.py
+++ b/Pilot/pilotTools.py
@@ -18,6 +18,7 @@ import threading
 from datetime import datetime
 from functools import partial, wraps
 from threading import RLock
+import warnings
 
 ############################
 # python 2 -> 3 "hacks"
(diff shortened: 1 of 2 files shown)


Pilot/pilotTools.py
```diff
@@ -214,53 +214,97 @@ def listdir(directory):
...
+        pilotReference = "sshtorque://" + ceName + "/" + batchSystemJobID.split(".")[0]
 
-    # Take the reference from the OAR batch system
+    # OAR
     if "OAR_JOBID" in os.environ:
-        flavour = "SSHOAR"
-        pilotReference = "sshoar://" + ceName + "/" + os.environ["OAR_JOBID"]
+        batchSystemType = "OAR"
+        batchSystemJobID = os.environ["OAR_JOBID"]
+
```
fstagni, line 254: `batchSystemParameters["BinaryPath"]` and ` batchSystemParameters["Queue"]` for this case?

Pilot/pilotTools.py
```diff
@@ -214,53 +214,97 @@ def listdir(directory):
...
+    batchSystemJobID = None
+    batchSystemParameters = {
+        "BinaryPath": "Unknown",
+        "Host": "Unknown",
+        "InfoPath": "Unknown",
+        "Queue": "Unknown",
+    }
+    # Flavour of the pilot
+    # Inform whether the pilot was sent through SSH+batch system or a CE
+    flavour = "Unknown"
```
fstagni, line 236: Why not anymore `flavour = "DIRAC"` ?

Pilot/pilotTools.py
```diff
@@ -214,53 +214,97 @@ def listdir(directory):
...
-def getFlavour(ceName):
+def getSubmitterInfo(ceName):
+    """
+    Check the environment variables to determine the type of batch system and CE used 
+    to submit the pilot being used and return this information in a tuple.
+    """
 
-    pilotReference = os.environ.get("DIRAC_PILOT_STAMP", "")
-    flavour = "DIRAC"
+    pilotReference = os.environ.get("DIRAC_PILOT_STAMP", "Unknown")
```
fstagni, line 223: Why from `""` to `"Unknown"` ?

Pilot/pilotTools.py
```diff
@@ -214,53 +214,97 @@ def listdir(directory):
     return contents
 
 
-def getFlavour(ceName):
+def getSubmitterInfo(ceName):
```
fstagni, line 217: Out if being conservative for the Pilot, please do not change this name.

Pilot/pilotTools.py
aldbr: I don't know, `"Unknown"` seemed more descriptive than `"DIRAC"` to me.
Any reason to name it `"DIRAC"`?

Pilot/pilotTools.py
```diff
@@ -214,53 +214,97 @@ def listdir(directory):
...
+        pilotReference = "sshoar://" + ceName + "/" + batchSystemJobID
 
     # Grid Engine
     if "JOB_ID" in os.environ and "SGE_TASK_ID" in os.environ:
-        flavour = "SSHGE"
-        pilotReference = "sshge://" + ceName + "/" + os.environ["JOB_ID"]
-    # Generic JOB_ID
-    elif "JOB_ID" in os.environ:
-        flavour = "Generic"
-        pilotReference = "generic://" + ceName + "/" + os.environ["JOB_ID"]
```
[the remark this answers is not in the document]
aldbr: Because I don't see where such a variable is set in the code (I might have just missed it).
If I understand correctly, this would happen if we use:
- `Local/SSH`CE + `Host` Batch system.
- `ARC` but no batch system under the hood (is it even possible?)

In the first case, `JOB_ID` does not seem to be set. There is `context.upper() + "_JOBID"` though.

Pilot/pilotTools.py
aldbr: Indeed, I will revert that.

Pilot/pilotTools.py
aldbr: If you want to be conservative, just keeping the same name if not enough as I am also changing the returned result...

I guess I would actually need to define 2 different methods: `getFlavour()`, which should be marked as deprecated and `getSubmitterInfo()`. What do you think?

aldbr: In my latest commit, I define both methods, `getFlavour()` is declared deprecated. I used `warnings`, which should be compatible from python2.6 to python3.11.
Any opinion?

fstagni: Perfect.

Pilot/pilotTools.py
fstagni: Because normally wherever we run through SSH we call the site "DIRAC.name.country"

Pilot/pilotTools.py
fstagni: Mmm maybe @atsareg knows what was this for. The feeling is that it could be dropped...

Pilot/pilotTools.py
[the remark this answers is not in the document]
fstagni: The first case (with `Host`) is rare but possible. Somehow that should be taken into consideration.

Pilot/pilotTools.py
aldbr: I reverted the line, but I still don't understand the logic here.

Pilot/pilotTools.py
aldbr: From what I have seen, the project is not very known but still exists and has active developers working on it.

Pilot/pilotTools.py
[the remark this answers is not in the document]
aldbr: So, `SSH` + `Host` was already taken into account, I just added the `Local` + `Host` case, which is going to replace `flavour = "Generic"`, unless you have a good reason to keep it.

fstagni (approved):

PR #225, opened 2023-12-08
Guess location of CVMFS deployed installation

fstagni:
This PR potentially makes it optional the specification of "preinstalledEnv" or "preinstalledEnvPrefix". 

If the deployment on CVMFS follows the "standard" setup of: `$location/{releaseProject}dirac/$version/$platorm/diracosrc` then adding `CVMFS_locations` list would be enough.

Loosely connected to https://github.com/DIRACGrid/DIRAC/pull/7359


fstagni:
All integration tests in https://github.com/fstagni/Pilot/actions/runs/7143604860.

PR #236, opened 2024-03-26
[devel] fix: use another environment variable to spot pilots running in htcondor

aldbr:
Pilots submitted to AREX and running through HTCondor under the hoods do not always have the `CONDOR_JOBID` variable set. This is not even part of the official documentation.

I picked up an environment variable from the documentation instead: https://htcondor.readthedocs.io/en/latest/users-manual/env-of-job.html#extra-environment-variables-htcondor-sets-for-jobs

From what I have observed so far, it is always present.
We should have less "`Unknown`" batch systems.

### DIFF
--- a/Pilot/pilotTools.py
+++ b/Pilot/pilotTools.py
@@ -271,24 +271,14 @@ def getSubmitterInfo(ceName):
         flavour = "SSH%s" % batchSystemType
         pilotReference = "sshslurm://" + ceName + "/" + batchSystemJobID
 
-    # Condor
-    if "CONDOR_JOBID" in os.environ:
+    # HTCondor
+    if "_CONDOR_JOB_AD" in os.environ:
         batchSystemType = "HTCondor"
-        batchSystemJobID = os.environ["CONDOR_JOBID"]
-        batchSystemParameters["InfoPath"] = os.environ.get("_CONDOR_JOB_AD", "Unknown")
+        batchSystemJobID = None # Not available in the environment
+        batchSystemParameters["InfoPath"] = os.environ["_CONDOR_JOB_AD"]
 
         flavour = "SSH%s" % batchSystemType
-        pilotReference = "sshcondor://" + ceName + "/" + batchSystemJobID
-
-    # # CEs/Batch Systems
-
-    # HTCondor
-    if "HTCONDOR_JOBID" in os.environ:
-        batchSystemType = "HTCondor"
-        batchSystemJobID = os.environ["HTCONDOR_JOBID"]
-
-        flavour = "HTCondorCE"
-        pilotReference = "htcondorce://" + ceName + "/" + batchSystemJobID
+        pilotReference = "sshcondor://" + ceName + "/" + os.environ.get("CONDOR_JOBID", pilotReference)
 
     # # Local/SSH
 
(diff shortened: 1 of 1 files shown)


Pilot/pilotTools.py
```diff
@@ -311,6 +301,11 @@ def getSubmitterInfo(ceName):
 
     # # CEs
 
+    # HTCondor
+    if "HTCONDOR_JOBID" in os.environ:
+        flavour = "HTCondorCE"
+        pilotReference = "htcondorce://" + ceName + "/" + batchSystemJobID
```
fstagni, line 307: Is `batchSystemJobID` going to be always `None` in this case? If that's the case, as it seems to be, apart that this as it is would raise a `TypeError`, we would end up anyway having a number of `pilotReference`s that are all the same.  If there's no `batchSystemID` identifiable, we better report `None` as `pilotReference`: in this case the `pilotUUID` would be used, which is guaranteed to be there and be unique.

Pilot/pilotTools.py
```diff
@@ -271,25 +271,15 @@ def getSubmitterInfo(ceName):
...
+    # HTCondor
+    if "_CONDOR_JOB_AD" in os.environ:
         batchSystemType = "HTCondor"
-        batchSystemJobID = os.environ["CONDOR_JOBID"]
-        batchSystemParameters["InfoPath"] = os.environ.get("_CONDOR_JOB_AD", "Unknown")
+        batchSystemJobID = None # Not available in the environment
+        batchSystemParameters["InfoPath"] = os.environ["_CONDOR_JOB_AD"]
 
         flavour = "SSH%s" % batchSystemType
         pilotReference = "sshcondor://" + ceName + "/" + batchSystemJobID
```
fstagni, line 281: Same problem here, `batchSystemJobID` would be `None`

aldbr: Oops sorry, I should have taken more time to read the code I wrote :sweat_smile: 
So I came up with a mixed solution, namely: if `CONDOR_JOBID` is defined (it seems to be the case when SSH+Condor is used), use it. Else use the pilot uuid.
I made the choice of keeping `sshcondor://ceName` as part of the pilot reference because it provides details about the CE and the Batch System used, but if you feel that only the pilot uuid should be used if `CONDOR_JOBID` is not set, I can modify the code.

Pilot/pilotTools.py
```diff
@@ -271,24 +271,14 @@ def getSubmitterInfo(ceName):
...
-
-        flavour = "SSH%s" % batchSystemType
-        pilotReference = "sshcondor://" + ceName + "/" + batchSystemJobID
-
-    # # CEs/Batch Systems
-
     # HTCondor
-    if "HTCONDOR_JOBID" in os.environ:
+    if "_CONDOR_JOB_AD" in os.environ:
         batchSystemType = "HTCondor"
```
fstagni, line 276: I think at this point it should be:

```
    # Condor
    if "_CONDOR_JOB_AD" in os.environ:
        batchSystemType = "Condor"
```

aldbr: Why? 
`Condor` is the old name for `HTCondor`: https://lists.cs.wisc.edu/archive/htcondor-users/2012-October/msg00110.shtml
I think it should be the opposite then, we should rename `Condor` as `HTCondor` in DIRAC.

fstagni (approved):
I merge this. It should be first tested in the certification setup before I merge `devel` to `master`.

PR #237, opened 2024-04-18
[master] Drop imp to support Python 3.12

chrisburr:
Totally untested, @fstagni feel free to take over as I probably won't get back to this until this evening


fstagni:
This will be for `devel`, and it's being properly tested in https://github.com/fstagni/Pilot/actions/runs/8736193657

PR #239, opened 2024-04-22
feat: adding to local cfg platform and CVMFS_locations

fstagni:


aldbr (commented):
This is a very minor and insignificant comment:

Since you are modifying the `README` file, what about converting it from `rst` to `md`?
The `README` file is rather simple, I don't think `rst` brings any benefit here.

PR #243, opened 2025-03-14
fix: Removed malformed and unused cfg

Robin-Van-de-Merghel:
This file (`PilotLoggerTest.cfg`) is not used at all in the code, and furthermore is a malformed file.

See (this documentation)[https://github.com/DIRACGrid/diraccfg/] about cfg files.


fstagni:
Yes, thanks! But please target `devel` branch (we should never push directly to `master`).

Robin-Van-de-Merghel:
Noted for next time!

Robin-Van-de-Merghel:
I've changed it to `devel` now!

fstagni (approved):

PR #245, opened 2025-03-14
Adding variables to `integration.yml`

Robin-Van-de-Merghel:
It would help if in the future this repository changes for a full DiracX use.

Adding variables to `integration.yml` would help if in the future this repository changes for a full DiracX use. 

It would also helps this repo to change with time. For example, with `USER_DN`, for whatever reason if the current hard-coded user leaves CERN, or if we want to add a custom `test` user, we could. 

Finally, it would also help contributers that forked this project, to tailor it easily for their proper use (for example to test it with their credentials, as well as with their own site).

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -40,17 +40,17 @@ jobs:
         run: |
           cd Pilot
           cp ../tests/CI/${{ matrix.pilot_schema }} pilot.json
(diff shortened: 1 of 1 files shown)


fstagni:
I think it's good for `CONFIGURATION_SERVER_URL` and `USER_DN`, for the others there's no real need.

PR #248, opened 2025-03-18
feat: Adding JWT support alongside X509 auth

Robin-Van-de-Merghel:
Added `BaseConnectedRequest` to support multiple auth mecanism. Derived from it are two classes:

- `TokenBasedRequest`
- `X509BasedRequest`

You can use either of these. 

> The code still work as previously with a X509 authentification, using `X509BasedRequest`, see [here](https://github.com/Robin-Van-de-Merghel/Pilot/blob/87913a182c8487cc8a4f67776ec9d6d423784100/Pilot/pilotTools.py#L674-L701)

Later, we can transition to a `TokenBasedRequest`. For that, we'll need to replace the `X509_USER_PROXY` env variable with a `JWT_USER` or something similar.


Pilot/pilotTools.py
```diff
@@ -19,6 +19,15 @@
...
 
+try:
+    from Pilot.proxyTools import (
+        X509BasedRequest
+    )
+except ImportError:
+    from proxyTools import (
+        X509BasedRequest
+    )
+
```
fstagni, line 30: This can be added below, where there is 
```
try:
    from Pilot.proxyTools import getVO
except ImportError:
    from proxyTools import getVO
```

simply doing 

```python

try:
    from Pilot.proxyTools import getVO, X509BasedRequest
except ImportError:
    from proxyTools import getVO, X509BasedRequest
```

Pilot/proxyTools.py
```diff
@@ -1,11 +1,27 @@
-""" few functions for dealing with proxies
-"""
+"""few functions for dealing with proxies and authentification"""
```
fstagni, line 1:
```suggestion
"""few functions for dealing with proxies and authentication"""
```

Pilot/proxyTools.py
```diff
@@ -1,11 +1,27 @@
...
 
 from __future__ import absolute_import, division, print_function
 
 import re
 from base64 import b16decode
 from subprocess import PIPE, Popen
+import ssl
+import sys
+import os
```
fstagni, line 10: Just order (what `isort` does)

Pilot/proxyTools.py
```diff
@@ -65,3 +82,96 @@ def getVO(proxy_data):
...
+    """Connected Request with JWT support"""
+
+    def __init__(self, url, caPath, jwtData):
+        super(TokenBasedRequest, self).__init__(url, caPath, "TokenBasedConnection")
+
+        self.jwtData = jwtData
+
+    def executeRequest(self, raw_data, headers={"User-Agent": "Dirac Pilot [Unknown ID]"}):
+        # Adds the JWT in the HTTP request (in the Bearer field)
+        headers["Bearer"] = self.jwtData
```
aldbr, line 151: I think it's: 
```suggestion
        headers["Authorization"] = f"Bearer: {self.jwtData}"
```

Pilot/proxyTools.py
```diff
@@ -65,3 +82,96 @@ def getVO(proxy_data):
...
+        """
+        return "Dirac Pilot [%s]" % pilotUUID
+
+    def _prepareRequest(self):
+        """As previously, loads the SSL certificates of the server (to avoid "unknown issuer")"""
+        # Load the SSL context
+        self._context = ssl.create_default_context()
+        self._context.load_verify_locations(capath=self.caPath)
+
+    def executeRequest(self, raw_data, headers={"User-Agent": "Dirac Pilot [Unknown ID]"}):
```
aldbr, line 117: I would move `headers` as a private attribute of `BaseConnectedRequest`, to avoid repeating the default value multiple times and also because it can be misleading:

```python
In [6]: def executeRequest(headers={"User-Agent": "Dirac Pilot [Unknown ID]"}, i=1):
   ...:     headers[f"Bearer{i}"] = "coucou"
   ...:     print(headers)
   ...: 

In [7]: executeRequest(i=1)
{'User-Agent': 'Dirac Pilot [Unknown ID]', 'Bearer1': 'coucou'}

In [8]: executeRequest(i=2)
{'User-Agent': 'Dirac Pilot [Unknown ID]', 'Bearer1': 'coucou', 'Bearer2': 'coucou'}

In [9]: executeRequest(i=3)
{'User-Agent': 'Dirac Pilot [Unknown ID]', 'Bearer1': 'coucou', 'Bearer2': 'coucou', 'Bearer3': 'coucou'}
```

(You are not impacted by this issue, it's just to highlight the potential risk.)

Also you could directly update the user agent value in the headers in `generateUserAgent()`

Pilot/proxyTools.py
```diff
@@ -54,3 +82,96 @@ def getVO(proxy_data):
...
+        """Execute a HTTP request with the data, headers, and the pre-defined data (SSL + auth)
+
+        Args:
+            raw_data (dict): Data to send
+            headers (dict, optional): Headers to send, helps to track requests. Defaults to {"User-Agent": "Dirac Pilot [Unknown ID]"}.
+
+        Returns:
+            str: Response of the HTTP request
```
aldbr, line 125: You are using Google-style docstrings while we are generally using the reST format.
It's an opinionated choice that could be discussed but it's better to be consistent throughout the code.

Robin-Van-de-Merghel: Corrected ✅

Pilot/proxyTools.py
Robin-Van-de-Merghel: I added header as a parameter to customize it easily. I do agree that it would be nice to put it as a "private" field.

In the future, if needed to add headers (for example, `Content-Type` or `Accept`), we could add a `conn.addHeader()`

Robin-Van-de-Merghel:
I am still testing the code. Need changes.

Pilot/proxyTools.py
aldbr: You can make it `protected` (may be not private because you need to access it from the inheriting class) by writing `self._headers` instead of `self.headers`.

aldbr:
(You need to fix the conflicts with `pilotTools`)

Robin-Van-de-Merghel:
Improved it so we can just add an `env` variable to acces it.

I would like a discussion about storing the JWT:

- an `env` with directly the encoded token
- an `env` with a path to the token (as for `certs`)
- a `string` parameter in the function, that `sendMessage` receives from `RemoteLogger` that himself has from the `pilotConfig`

For now I've chosen the first choice, tell me if you think it is right.

Robin-Van-de-Merghel:
[Issue with python 2]

Pilot/proxyTools.py
```diff
@@ -30,15 +48,10 @@ def findExtension(oid, lines):
...
-    Raises:
-        Exception: Any error related to openssl
-        NotImplementedError: Not documented error
-
-    Returns:
-        str: A VO
+    :param proxy_data: Bytes for the proxy chain
+    :type proxy_data: bytes
+    :return: A VO
+    :rtype: str
```
fstagni, line 54: Just a note: the above way of documenting is the popular "google style", and there's nothing wrong with it. In other parts of the code we used the Sphinx style.

Pilot/proxyTools.py
```diff
@@ -65,3 +78,120 @@ def getVO(proxy_data):
...
+        ctx = self._context  # Save in case of an insecure request
+
+        if insecure:
+            # DEBUG ONLY
+            # Overrides context
+            ctx = ssl.create_default_context()
+            ctx.check_hostname = False
+            ctx.verify_mode = ssl.CERT_NONE
+
+        if sys.version_info[0] >= 3:
```
fstagni, line 144:
```suggestion
        if sys.version_info.major >= 3:
```

Pilot/proxyTools.py
```diff
@@ -65,3 +78,120 @@ def getVO(proxy_data):
...
+            # Python 3 code
+            with urlopen(request, context=ctx) as res:
+                response_data = res.read().decode("utf-8")  # Decode response bytes
```
fstagni, line 147: Isn't it

```suggestion
                response_data = res.read()
```

?

Pilot/pilotTools.py
```diff
@@ -696,33 +696,61 @@ def sendMessage(url, pilotUUID, wnVO, method, rawMessage):
...
-    context.load_verify_locations(capath=caPath)
-
-    message = json.dumps((json.dumps(rawMessage), pilotUUID, wnVO))
+    message = json.dumps((json.dumps(rawMessage), pilotUUID, wnVO))    
+    raw_data = {"method": method, "args": message}
+
+    config = None
+    
+    if withJWT:
+        jwtData = os.getenv("JWT_TOKEN")
```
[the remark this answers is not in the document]
Robin-Van-de-Merghel: I could replace it with a dictionnary, it would make more sens I guess.

Pilot/proxyTools.py
Robin-Van-de-Merghel: Yes, I've started to use google style, but Alexandre (I don't remember where) to use redoc one, I'm correcting my old commit.

Pilot/proxyTools.py
Robin-Van-de-Merghel: Tested:

```
>>> # Python3
>>> type(res.read())
<class 'bytes'>


>>> # Python2 
>>> res = urlopen(request, context=ctx)
>>> type(res.read())
<type 'str'>
```

Robin-Van-de-Merghel:
Is there other changes to do?

Pilot/pilotTools.py
```diff
@@ -687,7 +687,7 @@ def cancelTimer(self):
             self._timer.cancel()
 
 
-def sendMessage(url, pilotUUID, wnVO, method, rawMessage):
+def sendMessage(url, pilotUUID, wnVO, method, rawMessage, withJWT=False, jwt={}):
```
fstagni, line 690: You can do it without the boolean flag

```suggestion
def sendMessage(url, pilotUUID, wnVO, method, rawMessage, jwt={}):
```

and later simply check `if jwt`

Pilot/pilotTools.py
```diff
@@ -696,33 +696,65 @@ def sendMessage(url, pilotUUID, wnVO, method, rawMessage):
...
+    config = BaseRequest(
+        "%s/api/auth/pilot-login%s" % (diracXURL, data_in_url),
+        caPath=caPath
+    )
+    
+    config.addHeader("Content-Type", "application/json")
+    
+    config.generateUserAgent(pilotUUID=pilotUUID)
+
+    # TODO: Add new cert for DiracX (different from Dirac)
```
fstagni, line 755: I do not understand this comment, can you explain?

Pilot/pilotTools.py
```diff
@@ -996,6 +1034,7 @@ def __init__(self):
             ("y:", "CEType=", "CE Type (normally InProcess)"),
             ("z", "pilotLogging", "Activate pilot logging system"),
             ("C:", "configurationServer=", "Configuration servers to use"),
+            ("", "diracXServer=", "DiracX server to use"),
```
fstagni, line 1037:
```suggestion
            ("", "diracx_URL=", "DiracX Server URL to use"),
```
just to be consistent with the DIRACX_URL env variable.

Pilot/proxyTools.py
```diff
@@ -65,3 +78,120 @@ def getVO(proxy_data):
...
+    
+    def addJwtToHeader(self):
+        # Adds the JWT in the HTTP request (in the Bearer field)
+        self.headers["Authorization"] = "Bearer: %s" % self.jwtData
+
+
+class X509BasedRequest(BaseRequest):
+    """Connected Request with X509 support"""
+
+    def __init__(self, url, caPath, certEnv):
```
fstagni, line 178: We need to support 2 cases here:
- the pilot's credential is a proxy -- this is the most common case, and I do not see it treated here
- the pilot's credential is a host certificate (cert+key pair). This is the "special case" and only to be considered when running with `useServerCertificate` flag true (this flag will be in the pilot parameters)

fstagni: For the 2nd case the location of the certificates to use is in the `certsLocation` pilot parameter flag. It's important to use this instead of being confused with the actual host certificate (the certificate of the host where the pilot is running)

Pilot/pilotTools.py
Robin-Van-de-Merghel: Dirac and DiracX are located on two different servers with two different domains. If a pilot can talk *directly* to DiracX without passing by Dirac, it can create a problem because we do non-insecured requests.

If we assume that a pilot can and will always talk to one Dirac/DiracX server, then there's no need for this.

Robin-Van-de-Merghel: I assume that Dirac and DiracX can have a different domain name, and that they can have a different server certificate.

We need a certificate to use properly `TLS/SSL` (verify that `dirac.com` is in fact DiracX/Dirac and not another random/malicious server). Currently we have this to make sure that `TLS/SSL` is used:

```
self._context = ssl.create_default_context()
self._context.load_verify_locations(capath=self.caPath)
```

Normally one certificate is associated with one domain (a cert for `domain1.dirac.com` is not valid for `domain2.dirac.com`, except if we have a certificate that accepts multiple domains, like with a [wildcard](https://www.ssl.com/article/what-is-a-wildcard-ssl-certificate/)). 

So, if DiracX and Dirac have a different domain, we could have an error with `ssl` saying that it can't verify the server because the pilot awaits a certificate saying "I'm Dirac" and the server announces "I'm DiracX".

Pilot/proxyTools.py
Robin-Van-de-Merghel: This is an old code that I adapted. But wasn't it already treated at the beginning?

On `PilotParams` we have the case for `useServerCertificate`:
```py
        if self.useServerCertificate:
            self.installEnv["X509_USER_PROXY"] = self.certsLocation
            os.environ["X509_USER_PROXY"] = self.certsLocation
```

If we don't set `self.useServerCertificate`, technically we search for a `X509_USER_PROXY` env variable that is already set, and that defines a `USER_PROXY` instead of a host certificate?

For what I understand, both cases are already treated? Or I did not understand well the beginning.

Robin-Van-de-Merghel:
Better with a command than a function that retrieves the JWT. Uses `dirac pilot-login <ref> <secret>` defined in [this PR](https://github.com/DIRACGrid/diracx/pull/421) of DiracX.

Robin-Van-de-Merghel:
Needs an update to match DiracX's PR.

Robin-Van-de-Merghel:
TODO: `self.pilotStamp = os.environ.get("DIRAC_PILOT_STAMP", self.pp.pilotUUID)`

Pilot/pilotTools.py
```diff
@@ -536,36 +534,49 @@ def __init__(
...
+        escaped = json.dumps(message)[1:-1]  # remove outer quotes
+
+        # Split on escaped newlines
+        splitted_message = escaped.split("\\n")
+
+        output = []
+        for mess in splitted_message:
+            if mess:
+                output.append({
+                    "timestamp": datetime.utcnow().strftime("%Y-%m-%dT%H:%M:%S.%fZ"),
```
Robin-Van-de-Merghel, line 552: Extract to the top

Pilot/pilotTools.py
```diff
@@ -974,7 +989,6 @@ def __init__(self):
             ("c", "cert", "Use server certificate instead of proxy"),
             ("d", "debug", "Set debug flag"),
             ("e:", "extraPackages=", "Extra packages to install (comma separated)"),
-            ("g:", "loggerURL=", "Remote Logger service URL"),
```
Robin-Van-de-Merghel, line 977: Avoid removing it to prevent bugs?

Pilot/pilotTools.py
```diff
@@ -1276,7 +1325,6 @@ def __initJSON2(self):
         if pilotLogLevel.lower() == "debug":
             self.debugFlag = True
         self.log.debug("JSON: Remote logging: %s" % self.pilotLogging)
-        self.log.debug("JSON: Remote logging URL: %s" % self.loggerURL)
```
Robin-Van-de-Merghel, line 1279: Replace with `diracxServer`

Pilot/dirac-pilot.py
```diff
@@ -124,3 +130,20 @@
...
             sys.exit(-1)
+
+    log.info("Pilot tasks finished.")
+   
+    if not remote:
+        log.buffer.flush()
+
+    if pilotParams.jwt:
+        if remote:
+            log.buffer.flush(force=True)
```
Robin-Van-de-Merghel, line 141: Remove `force` here, added later in #260

Issue #249, opened 2025-03-20
Send message has wrong number of arguments, leading to a crash


Robin-Van-de-Merghel:
In the [`logFinalizer` wrapper](https://github.com/DIRACGrid/Pilot/blob/aa95e04f33a62b04b203e1f3fbd47721cefc937e/Pilot/pilotCommands.py#L96), [`sendMessage`](https://github.com/DIRACGrid/Pilot/blame/aa95e04f33a62b04b203e1f3fbd47721cefc937e/Pilot/pilotTools.py#L665) (who hasn't changed for at least a year) is called with the wrong number of arguments, leading to a crash of pilots:

```
2025-03-20T12:29:47.514037Z DEBUG [LaunchAgent] Return code of dirac-agent WorkloadManagement/JobAgent -o MaxCycles=1 -o PollingTime=20 -o StopOnApplicationFailure=True -o StopAfterFailedMatches=10 -o LogLevel=DEBUG -s /Resources/Computing/CEDefaults -o WorkingDirectory=/pilot-tester -o /LocalSite/CPUTime=392400 -o /DIRAC/Security/UseServerCertificate=yes -o /LocalSite/InstancePath=/pilot-tester  -o /AgentJobRequirements/ExtraOptions=pilot.cfg --cfg pilot.cfg --cfg /pilot-tester/example.cfg: 0
2025-03-20T12:29:47.514992Z INFO [LaunchAgent] DiskSpace (MB) = 162783
2025-03-20T12:29:47.515132Z INFO [LaunchAgent] Flushing the remote logger buffer for pilot on sys.exit(): robin-s-debug-27743 (exit code:0)
2025-03-20T12:29:47.659655Z ERROR [LaunchAgent] Remote logger couldn't be finalised sendMessage() missing 1 required positional argument: 'rawMessage' 
Done.
```

This bug has been sleeping here for at least a year, maybe it impacted some pilots.

fstagni:
Fixed and deployed to `master`.

PR #250, opened 2025-03-20
Fix to logFinalizer: adding wnVO

Robin-Van-de-Merghel:
Related to #249 , added wnVO in `sendMessage` by adding to the remote logger field.

### DIFF
--- a/.github/workflows/basic.yml
+++ b/.github/workflows/basic.yml
@@ -5,7 +5,7 @@ on: [push, pull_request]
 jobs:
 
   shellcheck:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-24.04
     if: github.event_name != 'push' || github.repository == 'DIRACGrid/DIRAC'
     timeout-minutes: 10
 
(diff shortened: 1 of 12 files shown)


fstagni:
@martynia can you confirm this?

martynia:
Yes, it is clearly a bug. Thank you for pointing it out. We introduced it with #230. It affects the finaliser as a result of which log files are note renamed (.log added), but still accessible by the client, although most likely not in their complete form.

The pilots are not affected by the bug. If the sys.exit() code is not zero, pilot has crashed already ;-)
If the code is 0 [https://github.com/DIRACGrid/Pilot/blob/746d6b43ec852cb1d038a9f846348b11810eda5b/Pilot/pilotCommands.py#L1058](https://github.com/DIRACGrid/Pilot/blob/746d6b43ec852cb1d038a9f846348b11810eda5b/Pilot/pilotCommands.py#L1058)

then the pilot finishes fine (as shown in #249) but the file is not "finalised". 
Note: The finaliser relies on sys.exit(0) statement in the last command in the list of commands, otherwise the log will not be finalised.

fstagni (approved):

Robin-Van-de-Merghel:
@fstagni I did not know if you wanted to merge into `master` or into `devel` to fix it (while there is only one change in the `devel` branch)

fstagni:
always merging to `devel` first

PR #259, opened 2025-05-23
Robin add pilot logging

Robin-Van-de-Merghel:
Pilot based on #248 , linked with https://github.com/DIRACGrid/diracx/pull/550 to support logging.

It is working. Needs maybe test in workflows.


Robin-Van-de-Merghel:
Merged with #248

PR #260, opened 2025-06-27
Robin add pilot logging

Robin-Van-de-Merghel:
Splitted and based on #248

### DIFF
--- a/Pilot/dirac-pilot.py
+++ b/Pilot/dirac-pilot.py
@@ -69,25 +69,31 @@ if __name__ == "__main__":
     # print the buffer, so we have a "classic' logger back in sync.
     sys.stdout.write(bufContent)
     # now the remote logger.
-    remote = pilotParams.pilotLogging and (pilotParams.loggerURL is not None)
-    if remote:
+    remote = pilotParams.pilotLogging and (pilotParams.diracXServer is not None)
+    if remote and pilotParams.jwt:
         # In a remote logger enabled Dirac version we would have some classic logger content from a wrapper,
         # which we passed in:
         receivedContent = ""
         if not sys.stdin.isatty():
             receivedContent = sys.stdin.read()
         log = RemoteLogger(
-            pilotParams.loggerURL,
+            pilotParams.diracXServer,
             "Pilot",
             bufsize=pilotParams.loggerBufsize,
             pilotUUID=pilotParams.pilotUUID,
             debugFlag=pilotParams.debugFlag,
-            wnVO=pilotParams.wnVO,
+            jwt=pilotParams.jwt
         )
         log.info("Remote logger activated")
-        log.buffer.write(receivedContent)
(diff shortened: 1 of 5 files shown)


Pilot/dirac-pilot.py
```diff
@@ -63,25 +69,32 @@
...
-            pilotParams.loggerURL,
+            pilotParams.diracXServer,
             "Pilot",
             bufsize=pilotParams.loggerBufsize,
             pilotUUID=pilotParams.pilotUUID,
             debugFlag=pilotParams.debugFlag,
-            wnVO=pilotParams.wnVO,
+            jwt=pilotParams.jwt
         )
+        log.info(str(pilotParams.jwt))
```
Robin-Van-de-Merghel, line 87: leak

Pilot/pilotTools.py
```diff
@@ -622,34 +639,32 @@ def __init__(self, senderFunc, bufsize=1000, autoflush=10):
             self._timer.start()
         else:
             self._timer = None
-        self.output = StringIO()
```
martynia, line 625: There was an idea behind the buffer being a `StringIO `object. The buffer was first created in the pilot wrapper [(https://github.com/DIRACGrid/DIRAC/pull/6208/files#diff-bbbbde6222aeb320aa12b3df745daf44810996d2bf93c760e10625e1f6e4ed59)
](https://github.com/DIRACGrid/DIRAC/pull/6208/files#diff-bbbbde6222aeb320aa12b3df745daf44810996d2bf93c760e10625e1f6e4ed59) and passed to the pilot. A string appender was also added to a (classic) dirac logger). This served 2 purposes:

1. accumulate logs before the pilot was started.
2. enable to log from Dirac (not pilot) w/o any changes to Dirac code, which from now on would log to `stdout` and a string buffer.

I cannot tell at this level if your code allows remote logging from Dirac (and DiracX) using loggers they have.

Pilot/pilotCommands.py
```diff
@@ -92,16 +90,28 @@ def wrapper(self):
...
-                self.log.error("Remote logger couldn't be finalised %s " % str(exc))
+            if self.pp.jwt:
+                try:
+                    self.log.error(str(exCode))
+                    self.log.error(traceback.format_exc())
+                    self.log.buffer.flush(force=True)
+                except Exception as exc:
+                    self.log.error("Remote logger couldn't be finalised %s " % str(exc))
+            else:
+                # No force here because there's no remote logger if we're here
```
martynia, line 101: Not sure what you mean here. You are clearly within remote logger (if the remote logger is not enabled the wrapper returns at line 80). And the classic logger has no buffer...

Pilot/pilotTools.py
martynia: OK, so how does DiracX get hold of the buffer ? Dirac got it via a string appender to a global Logger object. Which DiracX PR is it ?

PR #265, opened 2025-08-25
fix: Installing multiple modules in InstallDIRAC command

chrisburr:


chrisburr:
Where are those documented? And I'm generally not keen to give certificates to GitHub

chaen:
we clearly shouldn't have our cert in github !

fstagni (approved):

PR #266, opened 2025-09-22
feat: removed python2 support

fstagni:


Untitled
aldbr: Is that expected? 😅

aldbr (approved):

chrisburr:
Why was this merged before removing the python 2 pilot tests from DIRAC?

Also https://github.com/DIRACGrid/Pilot/pull/266/commits/264587ce7287a9acd656842f29284b49958f2f9e is broken. The commit message talks about absolute imports when it's using relative ones and relative ones don't make sense in the way the pilotwrapper works.

PR #277, opened 2026-02-11
Avoid silent failure when checking the worker node: /etc/redhat-release

### LINKED ISSUE #274: [Bug]: silent failure when checking the worker node: `/etc/redhat-release` access can be restricted
### Search before creating an issue

- [x] I have searched existing issues and confirmed this is not a duplicate

### Bug Description

On some worker nodes (seen in a HPC), the pilot fails during the `CheckWorkerNode` step when attempting to read `/etc/redhat-release`.

The file exists but is not readable due to site-specific restrictions (permissions or security policies), which causes a `PermissionError` and makes the pilot exit early, even though the information is only informational/logging-related.

### Steps to Reproduce

1. Run the pilot on a worker node where `/etc/redhat-release` exists but is not readable by the pilot user.
2. The `CheckWorkerNode` command attempts to open `/etc/redhat-release`.
3. The pilot fails with a `PermissionError` during this step.

```python
Uname      = Linux nukwa-...
Host Name  = nukwa-01...
Host FQDN  = nukwa-01...
  File "/home/sec-constraints/wn.py", line 12, in <module>
    with open(fileName, "r") as f:
PermissionError: [Errno 13] Permission denied: '/etc/redhat-release'
```

### Expected Behavior

The pilot should not fail if OS release files are unreadable.

OS identification is informational and should be best-effort:
- If the file cannot be read, the pilot should continue running.
- OS details should be logged when available, skipped otherwise.

### Actual Behavior

The pilot exits during `CheckWorkerNode` when attempting to read `/etc/redhat-release`, even though the file is optional and not required for execution.

### Environment

_No response_

### Relevant Log Output

```shell

```

### Additional Context

https://github.com/DIRACGrid/Pilot/blob/10330ec8648f71db3fb12e31d7fb4d340b4c38b7/Pilot/pilotCommands.py#L141-L149

`/etc/redhat-release` is apparently legacy, distribution-specific file and may be restricted or absent on some systems.

A more robust and standardized alternative is `/etc/os-release` from what I understand: https://www.freedesktop.org/software/systemd/man/latest/os-release.html

It looks like it is supported by all modern (and even less modern) Linux distributions.

Based on [example4](https://www.freedesktop.org/software/systemd/man/latest/os-release.html) (python3.10) or [example5](https://www.freedesktop.org/software/systemd/man/latest/os-release.html) (still python2 support???), I think we should:
- Prefer reading `/etc/os-release` (best-effort, with exception handling)
- Then `/usr/lib/os-release`
- Never fail the pilot if OS release information cannot be read

fstagni:
closes #274

### DIFF
--- a/Pilot/pilotCommands.py
+++ b/Pilot/pilotCommands.py
@@ -115,15 +115,19 @@ class CheckWorkerNode(CommandBase):
         self.log.info("Host FQDN  = %s" % socket.getfqdn())
         self.log.info("WorkingDir = %s" % self.pp.workingDir)  # this could be different than rootPath
 
-        fileName = "/etc/redhat-release"
-        if os.path.exists(fileName):
-            with open(fileName, "r") as f:
-                self.log.info("RedHat Release = %s" % f.read().strip())
-
-        fileName = "/etc/lsb-release"
-        if os.path.isfile(fileName):
-            with open(fileName, "r") as f:
-                self.log.info("Linux release:\n%s" % f.read().strip())
+        for fileName in ["/etc/os-release", "/usr/lib/os-release"]:
+            if os.path.isfile(fileName):
+                try:
+                    with open(fileName, "r") as f:
+                        for line in f:
+                            line = line.strip()
+                            if line.startswith(("NAME=", "VERSION=", "PRETTY_NAME=")):
+                                self.log.info(
+                                    "OS Release = %s" % line.split("=", 1)[1].strip('"')
+                                )
+                    break
+                except (OSError, IOError):
+                    self.log.debug("Could not read %s" % fileName)
 
         fileName = "/proc/cpuinfo"
         if os.path.exists(fileName):
--- a/Pilot/tests/Test_Pilot.py
+++ b/Pilot/tests/Test_Pilot.py
@@ -5,6 +5,8 @@ import os
 import shutil
 import stat
 import sys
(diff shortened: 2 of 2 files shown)


Pilot/pilotCommands.py
```diff
@@ -115,15 +115,29 @@ def execute(self):
...
-                self.log.info("Linux release:\n%s" % f.read().strip())
+        for fileName in ["/etc/redhat-release", "/etc/lsb-release"]:
+            if os.path.isfile(fileName):
+                try:
+                    with open(fileName, "r") as f:
+                        fContent = f.read().strip()
+                        self.log.info("Linux release:\n%s" % fContent)
+                    break
+                except (OSError, IOError):
+                    self.log.debug("Could not read %s" % fileName)
```
aldbr: Any reason for checking these files?
Don't we get the same info from the `os-release` file?

fstagni: I kept for the moment just to understand if there's any added value, but I agree with you that they do not seem to provide any. I remove then.

aldbr:
What about adding a `pytest` test for the `CheckWorkerNode` pilot command?
- one test that make sure the log containing the content of `os-release` exist
- one test where `os-release` is unavailable
- one test where `os-release` access is blocked

aldbr (approved):

PR #278, opened 2026-02-25
Fix integration tests

fstagni:


fstagni:
Running in https://github.com/fstagni/Pilot/actions/runs/22398274420

If those are OK, we can finally get rid of py2 also on `master` branch

aldbr (approved):
All the tests are passing from what I can see

PR #279, opened 2026-02-25
New master 

### LINKED ISSUE #276: [Bug]: Dropping python2 support
### Search before creating an issue

- [x] I have searched existing issues and confirmed this is not a duplicate

### Bug Description

We currently have a PR to drop python2 support but we have issues with the `diracgrid` repo on CVMFS: https://github.com/fstagni/Pilot/actions/runs/21029011343

Having a python3-only Pilot would obviously ease the development and would allow a better integration of `diracx` (through `diracx-api`).

### Steps to Reproduce

_No response_

### Expected Behavior

_No response_

### Actual Behavior

_No response_

### Environment

_No response_

### Relevant Log Output

```shell

```

### Additional Context

_No response_

fstagni:
- remove py2 
-  avoid silent failure when checking /etc/redhat-release

closes #276


fstagni:
All integration tests succeeded in https://github.com/fstagni/Pilot/actions/runs/22400042649

Pilot/pilotTools.py
```diff
@@ -787,17 +768,6 @@ def executeAndGetOutput(self, cmd, environDict=None):
...
-                    # Ensure outChunk is unicode in Python 2
-                    if isinstance(outChunk, str):
-                        outChunk = outChunk.decode("utf-8")
-                    # Strip unicode replacement characters
-                    # Ensure correct type conversion in Python 2
-                    outChunk = str(outChunk.replace("\ufffd", ""))
-                    # Avoid potential str() issues in Py2
-                    outChunk = unicode(outChunk)  # pylint: disable=undefined-variable
-                else:
-                    outChunk = str(outChunk.replace("\ufffd", ""))  # Python 3: Ensure it's a string
```
aldbr, line 800: Is that expected?

Pilot/__init__.py
aldbr, line 1: Any reason for removing this `__init__` file?

Pilot/tests/Test_simplePilotLogger.py
```diff
@@ -1,25 +1,17 @@
...
-
-try:
-    from Pilot.pilotTools import CommandBase, Logger, PilotParams
-except ImportError:
-    from pilotTools import CommandBase, Logger, PilotParams
-
 import unittest
+from unittest.mock import patch
+
+sys.path.insert(0, os.getcwd() + "/Pilot")
```
aldbr, line 12: Potentially fragile no? Since it's just a test it's probably ok, but I am wondering why we need that?
May be you can add a comment above because it's not so common to see that

Pilot/pilotTools.py
fstagni: I don't think it was really adding anything useful.

Pilot/__init__.py
fstagni: It was not adding anything useful. This is not a package to be discovered.

Pilot/tests/Test_simplePilotLogger.py
fstagni: It is fragile, but again the Pilot is not a python package, it is just a bunch of file that we copy over.

aldbr (approved):

